PR #13, opened 2026-05-24
server: complete separated API origin serving

quentinnippert:
# PR Summary - Separated API Origin Serving

## Why

Pario now uses a clearer browser topology:

```txt
atlas.example.com -> Atlas UI shell
app.example.com   -> custom app UI shell
api.example.com   -> API, auth, WebSocket, docs
```

This keeps browser UI hosts public and lightweight while making the API origin the only
authenticated boundary. Sessions, CSRF, auth redirects, docs, and runtime WebSockets are owned by
the API server instead of being co-hosted or proxied through UI servers.

## What Changed

| Area | Change |
| --- | --- |
| Atlas | Moved the built-in UI into a dedicated `@pario/atlas` package with `createAtlasApp(...)`. |
| Custom apps | Renamed the custom app entrypoint to `createCustomApp(...)` and inject runtime API/auth config. |
| Server | Made `@pario/server` API-only for browser deployments: API, auth, docs, and WebSockets. |
| CLI | `pario dev` / `pario start` now orchestrate separate API, Atlas, and custom app servers. |
| Browser config | Added explicit local/public origin resolution for API, Atlas, and app surfaces. |
| Client bootstrap | Atlas and generated custom apps call the API origin with credentials and in-memory CSRF. |
| WebSockets | Browser WebSocket clients connect to the API origin, while custom app HMR stays on the app origin. |
| Legacy cleanup | Removed same-origin UI serving paths, fixed-audience fallbacks, and compatibility aliases. |

## Key Behavior

- UI hosts no longer serve `/api`, `/auth`, `/ws`, or `/docs`.
- `createParioServer(...)` always requires an explicit browser-origin policy.
- The API validates browser origins before accepting credentialed requests.
- Atlas and custom app shells redirect unauthenticated users to API sign-in with absolute `returnTo`.
- CSRF tokens are bootstrapped from `GET /api/auth/session`, not read from `document.cookie`.
- API-owned HTML routes such as `/docs` still use API-origin auth redirects.

## Validation

```bash
bun test packages/server/tests/
bun --filter @pario/server typecheck
bun run typecheck
bun run check
```


demattosanthony (approved):
LGTM 🔥🚀

PR #14, opened 2026-05-24
add sandbox contract and local sandbox provider

demattosanthony:
## Summary

Ports the sandbox-only slice from the old Pario sandboxes PR into sixb:

- adds the provider-agnostic `Sandbox` / `SandboxFactory` contract and sandbox error exports in `@pario/core`
- wires an optional `sandboxes` factory through `createPario`, `Pario`, `ParioInstance`, and `ParioRuntimeContext`
- adds the `@pario/sandboxes-local` workspace package with local command execution, timeout/abort handling, host env allowlisting, and `none` / Seatbelt / bwrap isolation backends
- adds provider-independent sandbox contract tests plus local sandbox unit/integration coverage

This intentionally excludes S4 provider changes, S4 workspace/dependency updates, and S4 config rendering from the original PR.

## Validation

- `bun test packages/core/tests/sandboxes sandboxes/local/tests`
- `bun --filter @pario/core typecheck`
- `bun --filter @pario/sandboxes-local typecheck`
- `bun --filter @pario/core build`
- `bun --filter @pario/sandboxes-local build`
- `bun run check`
- `bun run typecheck`
- `bun run test`

### DIFF
--- a/packages/core/src/sandboxes/sandbox.ts
+++ b/packages/core/src/sandboxes/sandbox.ts
@@ -1,0 +1,71 @@
+/**
+ * Provider-agnostic sandbox surface.
+ *
+ * A Sandbox is the agent's command execution target. Concrete providers
+ * implement this interface while consumers only depend on Sandbox and
+ * SandboxFactory.
+ */
+
+export type SandboxStatus = "running" | "stopped" | "failed"
+
+/**
+ * Per-command overrides. Each field overrides the matching sandbox-level
+ * default set at factory.create(...).
+ */
+export interface RunCommandOptions {
+  /** Overrides the sandbox-level workingDirectory for this call. */
+  readonly cwd?: string
+  /** Merged on top of sandbox-level env. Per-call wins on collision. */
+  readonly env?: Readonly<Record<string, string>>
+  /** Overrides the sandbox-level timeout, in milliseconds. */
+  readonly timeout?: number
+  /** Per-call only; there is no sandbox-level equivalent. */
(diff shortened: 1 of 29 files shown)


quentinnippert:
@demattosanthony i rebased this one, LGTM, we merge ?

packages/core/src/sandboxes/sandbox.ts
```diff
@@ -0,0 +1,90 @@
...
+  stop(): Promise<void>
+  /** Stop and reclaim any provider-side resources. Idempotent. */
+  destroy(): Promise<void>
+
+  /**
+   * Returns system-prompt-ready instructions describing this sandbox.
+   * Providers may return defaultSandboxInstructions(this) directly or compose
+   * it with provider-specific advice.
+   */
+  getInstructions(): string
```
demattosanthony: should we remove the prompting logic out of the sandbox implementations? Should this just be apart of the agent core runtime instructions?


same with this part as well: 

```
/**
 * Default body providers can return from getInstructions() if they have
 * nothing custom. Pure function with no side effects.
 */
export function defaultSandboxInstructions(sandbox: Sandbox): string {
  return [
    "You have access to a sandboxed shell.",
    "Run commands via the runCommand tool.",
    `Working directory: ${sandbox.workingDirectory}`,
  ].join("\n")
}
```

quentinnippert: Yeah makes sense, i pushed the updates

PR #19, opened 2026-05-25
server: expose global action requests

quentinnippert:
# Pull Request Summary

## What Changed

This PR updates Pario actions from an object-only method model to a request model that can support both global and object-scoped actions.

```ts
defineAction("createInvoice")
  .params(...)
  .run(...)
```

```ts
defineAction("markPaid")
  .on(Invoice)
  .params(...)
  .run(...)
```

## New Request Surface

Global and object-scoped requests now go through the canonical action namespace:

```ts
await pario.actions.request({
  actionId: "createInvoice",
  params,
})
```

```ts
await pario.actions.request({
  actionId: "markPaid",
  subject: {
    kind: "object",
    objectTypeId: "Invoice",
    primaryId: "invoice:1",
  },
  params,
})
```

The old top-level runtime helpers were removed:

```ts
pario.requestAction(...)
pario.requestActionAndWait(...)
```

Object ergonomics remain available and typed:

```ts
await pario.objects(Invoice).byId("invoice:1").requestAction({
  action: markPaid,
  params,
})
```

## Runtime Model

Action requests now carry an explicit subject:

```ts
type ActionSubject =
  | { kind: "none" }
  | { kind: "object"; objectTypeId: string; primaryId: string }
```

This subject is stored on action events and action runs, allowing global actions to exist without fake object identifiers while preserving object-scoped validation.

## Safety And Validation

- Global actions reject object subjects.
- Object actions require an object subject.
- Object subjects are checked against the action target and inheritance rules.
- Params are validated before emitting `action.requested`.
- Request-side validators run before the event is emitted.
- The action worker rejects forged object subjects before running handlers.

## Storage And Workers

- Action run storage now records `subject_kind`, `object_type_id`, and `primary_id`.
- SQLite and Postgres schemas were updated.
- The action worker can execute both global and object-scoped actions.
- Workflow action nodes remain object-scoped for this PR and reject global actions.
- The server action route now delegates to `pario.actions.request(...)`.

## Verification

```bash
bun --filter @pario/core typecheck
bun run typecheck
bun run check
bun test packages/server/tests/httpContract.test.ts packages/core/tests/actions.test.ts
```


packages/server/src/routes/actions.ts
```diff
@@ -1,45 +1,83 @@
...
-    },
-    {
-      params: ActionParamsSchema,
-      body: RequestActionBodySchema,
-      response: {
-        200: ActionRequestedResponseSchema,
-        400: ErrorResponseSchema,
-        404: ErrorResponseSchema,
+          return { success: true, runId }
+        } catch (error) {
```
demattosanthony, line 84: should we consolidate to the single post actions api route with optional subject instead of having two separate ones?

quentinnippert: done!

demattosanthony (approved):

PR #31, opened 2026-05-27
objects: add directional link listing

demattosanthony:
Problem

Object detail pages need to show what an object is related to. The existing object links endpoint only listed outgoing links, so incoming relationships were invisible unless the client scanned other objects and rebuilt the graph itself.

What changed

- Add a `direction` query option to object link listing.
- Support `outgoing`, `incoming`, and `both` in in-memory, SQLite, and Postgres storage.
- Regenerate the client contract and simplify Atlas to load relationships with one request.
- Render object links in a separate Atlas `Links` card below `Details`.

API

```http
GET /api/objects/{objectTypeId}/{objectId}/links?direction=both
```

Client usage

```ts
listObjectLinks({
  path: { objectTypeId, objectId },
  query: { direction: "both" },
})
```

Verification

- `bun run generate:client`
- `bun run typecheck`
- `bun run check`
- `bun run test`
- `bun run build`

### DIFF
--- a/packages/atlas/src/pages/ObjectDetailPage.tsx
+++ b/packages/atlas/src/pages/ObjectDetailPage.tsx
@@ -5,6 +5,7 @@ import type {
   TelemetryHistory,
   TelemetryProperty,
 } from "@pario/client"
+import { decodeObjectId, encodeObjectId } from "@pario/client"
 import {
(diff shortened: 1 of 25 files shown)


quentinnippert (approved):
clean

PR #32, opened 2026-05-27
docs: add standalone docs website

demattosanthony:
@quentinnippert Played around a bit with what we can do for wrapping our /docs markdown files into a website. Made a simple version to start and see whats possible. Take a look and see what you like

https://docs.sixb.ai/

all pages having markdown rendered versions as well: https://docs.sixb.ai/concepts/sync.md

let me know if you like this custom approach or we should make use of pre built framework like docusaurous or nextra

I also updated the docs content and tried to improve it a bit 

feel free to push some updates

----

## Summary

- Add a standalone `apps/docs` React/Bun docs site that renders the markdown docs as clean static pages.
- Publish matching `.md` files for each rendered docs route for LLM-friendly access.
- Rewrite the core docs pages with a simpler developer-focused style.
- Add a GitHub Pages workflow for `main`, this branch, and manual dispatch so we can test hosting before merging.

## Testing

- `bun run docs:typecheck`
- `bun run docs:build`
- `bun run check .github/workflows/docs.yml`
- `bun run check docs/concepts/ontology.md docs/concepts/datasets.md docs/concepts/sync.md docs/concepts/pipeline.md docs/concepts/projection.md docs/concepts/rules.md docs/concepts/workflows.md`

## Notes

- This branch is intentionally included in the Pages workflow so we can test deployment before merging.
- Before merging or after testing, we should remove `docs-website-beginnings` from the Pages workflow trigger if we only want `main` deployments.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -53,6 +53,8 @@ jobs:
 
       - name: Configure Pages
         uses: actions/configure-pages@v5
+        with:
+          enablement: true
 
       - name: Upload Pages artifact
         uses: actions/upload-pages-artifact@v4
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,72 @@
+name: Docs
+
+on:
+  push:
+    branches:
+      - main
+      - docs-website-beginnings
+    paths:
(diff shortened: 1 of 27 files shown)


quentinnippert (approved):
looks good, added some stuff

PR #34, opened 2026-05-28
cli: split production runtime roles

demattosanthony:
## Summary

Split production runtime execution into explicit one-role commands so production process managers can scale, restart, and isolate each runtime boundary independently — and keep each role's startup cheap and role-local so starting every role at once (e.g. via PM2) does not stampede shared infrastructure.

**Role split**
- Added `pario api`, `atlas`, `sentinel`, `app`, `scheduler`, `orchestrator`, `functions`, `rules`, and `worker <type>` commands.
- Removed `pario start` instead of keeping a CLI process supervisor; `pario dev` remains the local all-in-one command.
- Updated `pario build` and Atlas/Sentinel/custom app serving so production UI commands serve prebuilt assets, including custom `--outdir` builds.
- Externalized DuckLake from runtime config bundles so DuckDB native dependencies resolve from the provider package at runtime.
- Made Postgres storage migrations concurrency-safe with advisory locks, and tightened runtime provider cleanup for commands that load project runtimes.

**Connection budget (role startup)**
- Split `prepareParioRuntime` into `migrateRuntimeStorage` and `checkRuntimeLakeDefinitions`.
- Service commands (`api`, `scheduler`, `orchestrator`, `functions`, `rules`, `worker`, `worker-group`) no longer attach DuckLake or run storage migrations at startup. `pario dev` still does both in-process for local use.
- Added `pario lake check` as the single deploy-time validator that attaches the lake and checks every dataset definition for drift.
- Added `pario worker-group [types...]` to co-host queue workers in one process for deployments with a small Postgres/DuckLake connection budget (one lake pool instead of one per worker). With no positional types it starts every registered worker type.

## Why

Production deployments should not run API, browser UIs, scheduler, orchestrator, functions, rules, and queue workers inside one long-lived CLI process. Splitting them lets Docker, Kubernetes, systemd, and similar supervisors manage each role explicitly.

Keeping `pario start` created too much process-supervisor behavior inside the CLI: child lifecycle, signal handling, log fan-out, startup ordering, migration preflight, and partial failure cleanup. That boundary belongs to the deployment platform.

Once roles are separate processes, their *startup* behavior matters: the previous model ran a full lake-definition compatibility check and storage migrations before every role started. DuckDB and its Postgres extension run in-process per process, so if PM2 starts API, scheduler, orchestrator, functions, rules, and workers together, every role attaches DuckLake and opens its Postgres metadata pool at boot — a thundering herd against the catalog, even for roles that never touch lake data. Migrations had the same problem at a smaller scale. Moving those to explicit release steps keeps service startup cheap and role-local. (See https://duckdb.org/docs/current/core_extensions/postgres/connection_pool.)

Write paths still re-validate their target dataset through the lake provider's `createDataset` before committing, so dataset definition drift continues to fail clearly at job time even between deploys.

## Before / After

`pario start` behavior:

```diff
- pario start
- starts API, UI, runtime, and worker role commands as supervised child processes
+ removed
+ use `pario dev` for local all-in-one development
+ use explicit role commands for production process managers
```

Release order — run deployment checks as explicit steps, then start services:

```diff
- pario db migrate
- pario api            # each role also ran migrations + a full lake check at startup
- pario scheduler
- pario worker sync
+ pario build
+ pario db migrate     # required release step (roles no longer migrate at startup)
+ pario lake check     # required release step (roles no longer open the lake at startup)
+ pm2 start ecosystem.config.cjs
```

Production topologies — choose by connection budget:

```bash
# Scaled: one process per queue type, each with its own lake pool
pario api
pario scheduler
pario orchestrator
pario worker sync
pario worker pipeline
pario worker projection

# Constrained: co-host the queue workers in one process (one lake pool)
pario api
pario scheduler
pario orchestrator
pario worker-group sync pipeline projection
```

Custom build output directories resolve assets next to the built entry:

```diff
- --entry dist/pario.config.js -> dist/.pario/dist/atlas
- --entry dist/pario.config.js -> dist/.pario/dist/sentinel
- --entry dist/pario.config.js -> dist/.pario/dist/app
+ --entry dist/pario.config.js -> dist/atlas
+ --entry dist/pario.config.js -> dist/sentinel
+ --entry dist/pario.config.js -> dist/app
```

## Validation

```bash
bun test packages/cli/tests/
# includes: runtime.test.ts (split prep + dev still migrates),
#           lake.test.ts (pario lake check), prod-roles.test.ts
#           (roles start without migrating or touching the lake),
#           worker-group.test.ts, commands.test.ts, build.test.ts, init.test.ts
DATABASE_URL=postgres://postgres:postgres@localhost:5432/bacnet_example bun test ./storage/pg/tests/pg-storage-migrations.e2e.ts
bun --filter @pario/cli typecheck
bun --filter @pario/pg typecheck
bun run check
```


quentinnippert:
very clean separation

quentinnippert (approved):

PR #44, opened 2026-06-01
ducklake: optimize writes and add maintenance

demattosanthony:
This PR fixes the DuckLake storage bloat path described in the local spec.

Periodic snapshot syncs were rewriting dataset tables even when source rows were unchanged. DuckLake correctly preserved the old files for time travel, but because Sixb was not running DuckLake maintenance, object storage kept accumulating duplicate Parquet files.

DuckLake docs for this behavior:
- [Recommended maintenance](https://ducklake.select/docs/stable/duckdb/maintenance/recommended_maintenance): DuckLake does not delete old data files automatically; snapshots must be expired and files cleaned up.
- [Expire snapshots](https://ducklake.select/docs/stable/duckdb/maintenance/expire_snapshots): physical removal starts by expiring snapshots, and dry runs are supported.
- [Cleanup of files](https://ducklake.select/docs/stable/duckdb/maintenance/cleanup_of_files): expired-snapshot files and orphaned files are deleted with `ducklake_cleanup_old_files` and `ducklake_delete_orphaned_files`.
- [Checkpoint](https://ducklake.select/docs/stable/duckdb/maintenance/checkpoint): `CHECKPOINT` runs the DuckLake maintenance sequence and uses `expire_older_than` / `delete_older_than` options.

## Change 1: Skip unchanged snapshot writes

Snapshot writes now compare the incoming source relation with the current target table before doing the replacement.

If row count and order-independent fingerprints match, the commit returns the existing latest dataset version and skips the `DELETE` / `INSERT` pair. Append mode is unchanged.

```ts
if (sameFingerprint(source, target)) {
  return {
    dataChangeExpected: false,
    sourceRowCount: source.rowCount,
    resultingRowCount: { kind: "exact", value: target.rowCount },
  }
}
```

This stops future duplicate snapshots for unchanged periodic syncs without adding row materialization or an expensive diff to the hot path.

## Change 2: Add provider-owned DuckLake maintenance

`DuckLakeStorage` now exposes `runMaintenance()` for DuckLake snapshot/file cleanup.

It supports:
- dry-run reporting
- snapshot expiration
- old-file cleanup
- orphan-file cleanup
- checkpoint execution
- read-only protection

Maintenance is intentionally **operator-driven**, not auto-scheduled. There is no provider-registered cron function; Sixb does not inject scheduled functions into a project's runtime. Operators run cleanup on demand (see Change 3), or call `runMaintenance()` directly. This keeps the lake-storage contract free of a scheduling hook and leaves cadence/orchestration to the operator.

## Change 3: Add `sixb lake cleanup`

Operators run maintenance directly from the CLI:

```bash
sixb lake cleanup --dry-run
sixb lake cleanup --expire-older-than "1 hour" --delete-older-than "1 hour"
sixb lake cleanup --expire-older-than "7 days" --delete-older-than "7 days"
```

The command is capability-based: it loads the configured runtime, checks whether the lake provider exposes `runMaintenance`, renders the report, and closes providers on success or failure.

DuckLake retention options persist, so the CLI prints a reminder after aggressive non-dry-run cleanup to restore the normal 7-day retention window.


demattosanthony:
@quentinnippert pushed the change

PR #46, opened 2026-06-04
Optimize latest run loading for list routes

quentinnippert:
# Bulk Latest Run Loading

## Summary

This PR removes the N+1 latest-run lookup from list routes for syncs, pipelines, and workflows.

Previously, each listed definition loaded its `latestRun` by calling the paginated run history API with
`limit: 1`, which also triggered a `COUNT(*)`. For a page with 13 syncs, this meant roughly 26 SQL
queries just to render latest-run badges.

## Before / After

| Route | Before | After |
| --- | --- | --- |
| `GET /api/syncs` | 1 latest-run query pair per sync | 1 bulk latest-run query |
| `GET /api/pipelines` | 1 latest-run query pair per pipeline | 1 bulk latest-run query |
| `GET /api/workflows` | 1 latest-run query pair per workflow | 1 bulk latest-run query |

## What Changed

- Added required bulk latest-run storage methods for sync, pipeline, and workflow runs.
- Implemented the methods in in-memory, SQLite, and Postgres storage.
- Updated list/detail routes to call the bulk latest-run APIs directly.
- Kept HTTP response shapes unchanged, including the existing `latestRun` field.
- Added regression coverage for route-level bulk loading and storage-level latest-run selection.

## Storage Notes

- Postgres uses `DISTINCT ON (...)` ordered by `started_at DESC, id DESC`.
- SQLite uses `ROW_NUMBER() OVER (PARTITION BY ...)` with the same ordering.
- In-memory storage uses a shared helper with matching ordering semantics.

## Validation

- `bun run typecheck`
- `bun run check`
- Targeted core, server, SQLite, Postgres, and sync-worker tests


demattosanthony (approved):
clean 🧼

PR #47, opened 2026-06-06
Speed up DuckLake version metadata reads

quentinnippert:
# Fast DuckLake Version History

## Summary

This PR makes DuckLake dataset version metadata reads faster and safer for Atlas dataset detail pages.

## What Changed

| Area | Before | After |
| --- | --- | --- |
| Historical schema | `DESCRIBE SELECT * FROM table AT VERSION` | Rebuilt from DuckLake catalog metadata |
| Version row counts | Fallback `COUNT(*) AT VERSION` | Metadata-only, omitted when unknown |
| `/versions?limit=1` | Could still trigger expensive historical table reads | Stays on metadata reads |

## Details

- Reconstruct historical schemas from `ducklake_column` using `begin_snapshot` / `end_snapshot`.
- Carry the DuckLake `table_id` through snapshot hydration so schema lookup targets the exact table metadata.
- Stop deriving missing `rowCount` during version hydration; row counts are returned only when already stored in SixB commit metadata.
- Add a regression test that fails if version listing reintroduces historical table scans.

## Validation

- `bun --filter @sixb/ducklake typecheck`
- `bun --filter @sixb/ducklake test:e2e:local`
- `bun --filter @sixb/ducklake test:e2e:remote`
- `bun run typecheck`
- `bun run check`

### DIFF
--- a/storage/ducklake/src/internal/ducklake-dataset-catalog.ts
+++ b/storage/ducklake/src/internal/ducklake-dataset-catalog.ts
@@ -7,18 +7,23 @@ import type {
 } from "@sixb/core"
 import { LakeStorageError, planDatasetDefinitionUpdate } from "@sixb/core"
 import type { DuckLakeStorageOptions } from "../types"
-import { getString } from "./duckdb-row"
+import { getBigIntLike, getBoolean, getOptionalBigIntLike, getString } from "./duckdb-row"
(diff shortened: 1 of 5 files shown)


demattosanthony (approved):

demattosanthony:
⚡️⚡️⚡️

PR #52, opened 2026-06-07
core: add ontology query metadata

demattosanthony:
Stack position: 1 of 6. This PR is based on `projection-link-cleanup` / #48 because the query-layer work was built on top of that branch.

## What changed

This adds ontology-level query/search metadata and validates it when the `OntologyRegistry` starts. The goal is to make queryability explicit before later PRs add the query IR and execution path.

Properties can now opt into query surfaces:

```ts
prop("name", "string", {
  required: true,
  query: {
    searchable: true,
    text: true,
    exact: true,
    sortable: true,
    weight: 5,
  },
})
```

Object types can define a search profile:

```ts
defineObjectType({
  id: "Customer",
  name: "Customer",
  properties: [
    prop("id", "string", { required: true, primary: true }),
    prop("name", "string", { query: { searchable: true, text: true } }),
  ],
  search: {
    title: "name",
    defaultText: ["name"],
    exact: ["id"],
  },
})
```

## Review notes

- `query.searchable: true` is the gate for all specific query flags.
- Text search is limited to static string-like fields.
- Vector metadata requires numeric arrays.
- Primary ids can appear in `search.exact` without explicit `query.exact` metadata.
- This PR intentionally does not add the object-query IR, runtime API, storage pushdown, HTTP routes, or docs.

## Verification

```bash
bun test packages/core/tests/define.test.ts packages/core/tests/validation.test.ts
bun --filter @sixb/core typecheck
```

Both commands passed locally.


quentinnippert (approved):
cleannnnn

PR #71, opened 2026-06-14
actions: add atomic edit commit storage

quentinnippert:
# PR Summary: Storage Transaction Foundation

## What Changed

- Adds the shared `storage.transaction(...)` contract.
- Adds guarded transaction storage: a `tx` cannot be reused after commit/rollback.
- Rejects nested storage transactions for now.
- Implements provider-level transactions for:
  - `InMemoryStorage`
  - `SqliteStorage`
  - `PostgresStorage`

```ts
await storage.transaction(async (tx) => {
  await tx.objects.applyObjectUpserted(...)
  await tx.actionRuns.recordCommit(...)
})
```

## Provider Behavior

```txt
InMemory
  snapshot before callback
  keep changes on success
  restore snapshot on failure

SQLite
  shared connection across all stores
  BEGIN IMMEDIATE / COMMIT / ROLLBACK

Postgres
  sql.begin(...)
  stores accept pool or transaction-bound client
  internal store transactions reuse the active transaction
```

## Why

This gives action execution a real storage-level commit boundary before moving edit commits into
the action worker transaction flow.

## Tests

- Adds transaction contract tests for in-memory, SQLite, and Postgres.
- Covers commit, rollback, nested transaction rejection, and escaped `tx` usage.
- Updates existing worker/example tests to use complete storage facades.

## Verification

```txt
bun run typecheck
bun run check
bun test packages/core/tests/storage-transaction.test.ts storage/sqlite/tests/sqlite-storage-transaction.test.ts
bun test packages/action-worker/tests/worker.test.ts
bun test packages/rules-worker/tests/worker.test.ts
bun test examples/acme-corp/tests/client-query.test.ts
bun --filter @sixb/pg test:e2e -- pg-storage-transaction.e2e.ts
```


demattosanthony:
Something feels off to me about having `EditStorage`. I would expect that to store edits, but it looks more like a transactional edit commit coordinator: it applies the edit plan to object/link storage and records the resulting diff on the action run.

I think it exists because those writes need to happen in one storage transaction. But maybe the cleaner abstraction is a unit-of-work API on storage instead of a separate `edits` store:

  ```ts
  export interface Storage {
    objects?: ObjectStorage
    actionRuns?: ActionRunStorage

    transaction?<T>(fn: (tx: Storage) => Promise<T>): Promise<T>
  }
```

 Then edit commit orchestration could live in core and use tx-bound storage implementations:

```ts
  await storage.transaction(async (tx) => {
    const run = await tx.actionRuns!.getById(...)

    const plan = await planEditBatch({
      projectId,
      ontology,
      storage: { objects: tx.objects! },
      batch,
    })

    await tx.objects!.applyEditCommitPlan(projectId, plan, committedAt)
    await tx.actionRuns!.recordCommit({
      id: runId,
      projectId,
      committedAt,
      diff: plan.diff,
    })
  })
```

Not sure this needs to block this PR though. We could merge the current patch and revisit later if it makes sense

demattosanthony (approved):

PR #74, opened 2026-06-16
Add scoped authorization layer

demattosanthony:
## Review ask

This is a draft PR and I would really like another set of eyes on the direction here.

Apologies for the size of the change. I can break this up later if that is the right move, but at this point it feels more useful to get feedback on whether the overall authorization shape is on the right track before splitting it into smaller mechanical pieces.

The main questions I would like feedback on:

- Is `sixb.as(context)` returning a small scoped SDK surface the right core boundary?
- Are groups -> roles -> grants the right base model for v1?
- Is the grant vocabulary expressive enough without becoming too magical?
- Are the enforced surfaces the right first slice?
- Does the auth example make the intended DX clear enough?

## Summary

Adds the first framework-level authorization layer on top of the existing authentication system.

The current auth system gives us authenticated users, sessions, invitations, and group membership. This PR adds the first authorization kernel: declarative groups, roles, grants, per-request authorization context resolution, and a scoped runtime returned by `sixb.as(context)` that only exposes operations we can enforce end to end.

This is intentionally not a full policy/marking/object-instance authorization system yet. The goal here is the base layer that lets the framework answer: who can see which object types, who can apply which actions, who can run which workflows, and who can read events.

### Layer 1: groups

Groups are named membership buckets. Auth strategies and invitations can place users into groups.

```ts
export const teamMembers = defineGroup("team-members")
export const securityAdmins = defineGroup("security-admins")
```

### Layer 2: roles

Roles are assigned to groups. A principal receives a role when one of their session group ids matches the role's `grantedTo` groups.

```ts
export const teamMemberAccess = defineRole("team-member.atlas-access", {
  grantedTo: [teamMembers],
  grants: [
    can.view(Note),
    can.apply(acknowledgeNote),
  ],
})
```

### Layer 3: grants

Grants are plain data. They are validated at startup, deduped by stable key, and resolved into fast per-request decision indexes.

Concrete grants:

```ts
can.view(Note)
can.apply(acknowledgeNote)
can.run(runAccessReview)
can.viewEvents()
```

Wildcard grants:

```ts
can.viewAllObjects()
can.applyAllActions()
can.runAllWorkflows()
can.viewEvents()
```

The auth example uses wildcard grants for the admin role:

```ts
export const securityAdminFullAccess = defineRole("security-admin.full-access", {
  grantedTo: [securityAdmins],
  grants: [
    can.viewAllObjects(),
    can.applyAllActions(),
    can.runAllWorkflows(),
    can.viewEvents(),
  ],
})
```

### Layer 4: authorization context

For each authenticated request, the server derives an authorization context once:

```ts
const authz = sixb.auth.contextFromSession(auth.session)
const scoped = sixb.as(authz)
```

The context stores:

- principal
- session id
- group ids
- matched role ids
- viewable object type ids
- applicable action ids
- runnable workflow ids
- wildcard grant booleans
- event-stream access boolean

Concrete `can.view(ObjectType)` grants are expanded to include subtypes during context resolution, so hot-path checks stay simple set lookups.

### Layer 5: scoped SDK

`sixb.as(context)` returns `ScopedSixb`, not another full `Sixb` implementation. This is meant to be a thin, allowlisted runtime surface over the same underlying core services.

The scoped surface currently exposes only operations with grant semantics that are enforced end to end:

```ts
scoped.list(...)
scoped.getObject(...)
scoped.objects(Note).list()
scoped.objects(Note).query().list()
scoped.requestAction(...)
scoped.runWorkflow(...)
scoped.readEvents(...)
```

It intentionally does not expose privileged surfaces like raw storage, auth runtime, writes, links, telemetry writes, providers, or registry mutation.

### Objects

- Broad object listings are narrowed to viewable object types.
- Explicit forbidden object type filters return `403`.
- Object identity reads hide forbidden objects as `404` to avoid existence leaks.
- Typed object sets (`sixb.as(ctx).objects(Type)`) enforce `can.view(Type)`.
- Object query, count, exists, and facets enforce view access for every object type touched by the query.
- Scoped object writes, links, telemetry writes, and link mutation fail closed because we do not have write/link grants yet.

### Ontology metadata

- `/api/object-types` only returns object types visible to the principal.
- `/api/object-types/:id` returns `404` for forbidden object types.
- Object type action metadata only includes actions that the principal can apply and, for object-bound actions, whose object type the principal can view.

This matters for Atlas because it prevents the UI from discovering hidden object types and avoids launching forbidden preview/count queries for those types.

### Actions

- `/api/actions` is narrowed to applicable actions.
- `/api/actions/:id` hides ungranted actions as `404`.
- Requesting an action requires `can.apply(action)`.
- Object-bound action requests also require `can.view(subjectObjectType)`.
- Action metadata follows the same rule: apply is not enough to discover an action bound to a hidden object type.

### Workflows

- Starting a workflow requires `can.run(workflow)` or `can.runAllWorkflows()`.
- Existing workflow metadata/history routes are not fully authorization-modeled yet. See out-of-scope section below.

### Events

- Reading `/api/events` requires `can.viewEvents()`.
- The event websocket requires `can.viewEvents()`.

Events are intentionally coarse for now because the stream can contain object payloads that bypass object-level read checks.

### Server request integration

The server now derives auth state once per request and exposes:

```ts
{ auth, authz, scoped }
```

Routes that have enforceable grant semantics use `scoped` rather than manually duplicating auth checks.

## Auth example

The auth example now works as a small authorization smoke test instead of only proving sign-in.

It defines:

- `Note`: visible to team members and admins
- `AdminNote`: visible to admins only
- `AccessRequest`: visible to admins only
- `acknowledge-note`: action team members can apply
- `resolve-access-request`: admin action
- `run-access-review`: admin workflow

It seeds one object of each type so Atlas is testable immediately after startup.

Expected behavior:

- team member sees only `Note`
- team member can apply `acknowledge-note`
- team member cannot see `AdminNote` or `AccessRequest`
- team member cannot apply admin actions, run admin workflows, or read events
- security admin can view all auth example objects, apply all actions, run all workflows, and read events through wildcard grants

## Generated client

OpenAPI/client generated artifacts were regenerated because several API routes now document `403` responses for authorization failures.

## What this version intentionally does not include

This PR does not implement every authorization case in the framework. The following are intentionally out of scope for this first layer:

- object instance policies / row-level object security
- property-level security, markings, or PII redaction
- `can.write(...)` or write grants
- link read/write grants
- telemetry read/write grants
- dataset grants
- sync grants
- connector grants
- pipeline grants
- projection grants
- rule grants
- webhook/webhook-run grants
- workflow run history visibility grants
- intervention review/submit authorization grants
- auth admin grants such as managing users, sessions, groups, or invitations
- a global `can.admin()` bypass

Some server routes for those resource families still behave as authenticated application surfaces rather than fully grant-scoped authorization surfaces. The scoped SDK fails closed for unsupported privileged operations, but full route coverage should be follow-up work with explicit grant semantics instead of implicit admin behavior.

## Why this shape

The goal is to keep the core SDK simple and safe:

- `Sixb` remains the privileged runtime for workers, syncs, server internals, setup, tests, and seed code.
- `sixb.as(context)` gives request handlers and app code a principal-scoped SDK.
- Enforcement lives in shared core services and leaf operations, not only in HTTP route wrappers.
- `ScopedSixb` is intentionally a small allowlist facade, not a second full `Sixb` implementation.
- Unsupported scoped operations fail closed until they have real grants.

This should let us add future grant families without changing the top-level programming model.

## Validation

Ran:

```bash
bun run generate:client
bun run typecheck
bun run build
bun run test
bun run check
git diff --check
```

Results:

- full typecheck passed
- full build passed
- full test suite passed: `1339 pass, 0 fail`
- Biome check passed
- generated client artifacts are up to date

## Follow-up recommendation

If the direction looks right, I would split follow-up work by grant family rather than trying to add every resource type here:

1. workflow run/intervention visibility and mutation grants
2. dataset/sync/pipeline/projection grants
3. write/link/telemetry grants
4. auth admin grants
5. object instance and property-level policy system

Again, apologies for the size of this one. I am opening it as draft because I mainly want feedback on whether the architecture and DX are right before polishing or splitting further.

### DIFF
--- a/packages/core/src/authorization/types.ts
+++ b/packages/core/src/authorization/types.ts
@@ -14,8 +14,8 @@ export interface GrantIndex {
   readonly objectTypes: { readonly view: ReadonlySet<string> }
   /** Action ids this principal may request. */
   readonly actions: { readonly apply: ReadonlySet<string> }
-  /** Workflow ids this principal may start. */
-  readonly workflows: { readonly start: ReadonlySet<string> }
+  /** Workflow ids this principal may run. */
+  readonly workflows: { readonly run: ReadonlySet<string> }
 }
 
 /** A role with its grants pre-expanded to concrete id sets at startup. */
--- a/packages/core/src/authorization/types.ts
+++ b/packages/core/src/authorization/types.ts
@@ -1,0 +1,35 @@
+/**
+ * Authorization context types.
+ *
+ * The context is plain, immutable data resolved once per request
+ * (`principal -> groups -> roles -> grants`). Grants resolve eagerly into set
+ * lookups so enforcement on the hot path is synchronous and allocation-free.
+ * Group and role ids are retained so decisions stay explainable.
+ */
+
+import type { Principal } from "../auth/types"
+
+export interface GrantIndex {
+  /** Object type ids viewable, expanded from grants (broad grants + subtypes). */
+  readonly objectTypes: { readonly view: ReadonlySet<string> }
+  /** Action ids this principal may request. */
+  readonly actions: { readonly apply: ReadonlySet<string> }
+  /** Workflow ids this principal may start. */
+  readonly workflows: { readonly start: ReadonlySet<string> }
+}
+
+/** A role with its grants pre-expanded to concrete id sets at startup. */
+export interface ResolvedRole {
+  readonly id: string
+  readonly grantedToGroupIds: readonly string[]
+  readonly grants: GrantIndex
+}
+
+export interface AuthorizationContext {
+  readonly principal: Principal
+  readonly sessionId?: string
+  readonly groupIds: readonly string[]
+  /** Roles whose grantedTo groups intersect the principal's memberships. */
+  readonly roleIds: readonly string[]
+  readonly grants: GrantIndex
+}
--- a/packages/core/src/objects/sdk/object-set.ts
+++ b/packages/core/src/objects/sdk/object-set.ts
@@ -4,16 +4,18 @@
  * while preserving compile-time property types inferred from the ontology definition.
  */
 
-import type { ActionDefinition, ActionRegistry } from "../../actions"
-import type { BlobStorage } from "../../blob-storage"
-import type { EventsRuntime } from "../../events"
-import type { LakeStorage } from "../../lake-storage"
-import type { OntologyRegistry, ValueType } from "../../ontology"
+import type { ActionDefinition } from "../../actions"
+import type { AuthorizationContext } from "../../authorization"
+import { assertAuthorized } from "../../authorization"
+import type { ValueType } from "../../ontology"
 import { OntologyValidationError } from "../../ontology/errors"
 import type { ObjectTypeWithPropertyTokens } from "../../ontology/tokens"
-import type { Queues } from "../../queues"
-import type { ObjectSet, ObjectSetListInput, TwinObject } from "../../runtime/types"
-import type { Storage } from "../../storage"
+import type {
+  ObjectSet,
+  ObjectSetListInput,
+  SixbRuntimeContext,
+  TwinObject,
+} from "../../runtime/types"
 import {
   requestActionAndWait as requestActionAndWaitLeaf,
   requestAction as requestActionLeaf,
@@ -31,19 +33,20 @@ export function createObjectSet<
   TObjectType extends ObjectTypeWithPropertyTokens,
   TRegisteredObjectTypes extends ObjectTypeWithPropertyTokens,
   TValueTypes extends readonly ValueType[],
->(params: {
-  objectType: TObjectType
-  projectId: string
-  ontology: OntologyRegistry
-  actionRegistry: ActionRegistry
-  events: EventsRuntime
-  lakeStorage: LakeStorage
-  blobStorage: BlobStorage
-  storage: Storage
-  queues: Queues
-}): ObjectSet<TObjectType, TValueTypes, TRegisteredObjectTypes> {
+>(
+  params: SixbRuntimeContext & {
+    readonly objectType: TObjectType
+    readonly authorization?: AuthorizationContext
+  }
+): ObjectSet<TObjectType, TValueTypes, TRegisteredObjectTypes> {
+  const { objectType } = params
+  const primaryProp = objectType.properties.find((p) => p.primary)
+  if (!primaryProp) {
+    throw new OntologyValidationError(`Object type '${objectType.id}' has no primary property`)
+  }
+  const primaryPropertyId = primaryProp.id
+
   const {
-    objectType,
     projectId,
     ontology,
     actionRegistry,
@@ -52,15 +55,9 @@ export function createObjectSet<
     blobStorage,
     storage,
     queues,
+    authorization,
   } = params
-
-  const primaryProp = objectType.properties.find((p) => p.primary)
-  if (!primaryProp) {
-    throw new OntologyValidationError(`Object type '${objectType.id}' has no primary property`)
-  }
-  const primaryPropertyId = primaryProp.id
-
-  const queryExecutor = createRuntimeQueryExecutor({ projectId, ontology, storage })
+  const queryExecutor = createRuntimeQueryExecutor({ projectId, ontology, storage, authorization })
 
   const resolvedCtx: ResolvedObjectContext = {
     projectId,
@@ -71,12 +68,14 @@ export function createObjectSet<
     blobStorage,
     storage,
     queues,
+    authorization,
     objectType,
     primaryPropertyId,
   }
 
   const objectSet = {
     get: async (id: string) => {
+      assertAuthorized(resolvedCtx, { kind: "object.view", objectTypeId: objectType.id })
       const row = await storage.objects.getByPrimaryId({
         projectId,
         objectTypeId: objectType.id,
@@ -105,6 +104,7 @@ export function createObjectSet<
     byId: (id: string) => createObjectByIdHandle<TObjectType, TValueTypes>(resolvedCtx, id),
 
     list: async (input?: ObjectSetListInput) => {
+      assertAuthorized(resolvedCtx, { kind: "object.view", objectTypeId: objectType.id })
       const result = await storage.objects.list({
         projectId,
         objectTypeId: objectType.id,
--- a/packages/core/src/security/types.ts
+++ b/packages/core/src/security/types.ts
@@ -1,8 +1,51 @@
+import type { ResolvedRole } from "../authorization/types"
+
 export interface GroupDefinition<TId extends string = string> {
   readonly kind: "group"
   readonly id: TId
   readonly label?: string
   readonly description?: string
+}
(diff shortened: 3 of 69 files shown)


packages/core/src/security/types.ts
```diff
@@ -5,6 +5,62 @@ export interface GroupDefinition<TId extends string = string> {
...
+  | ViewEventsGrant
+
+export interface RoleDefinition<TId extends string = string> {
+  readonly kind: "role"
+  readonly id: TId
+  readonly label?: string
+  readonly description?: string
+  readonly grantedToGroupIds: readonly string[]
+  readonly grants: readonly GrantDefinition[]
+}
```
quentinnippert, line 49: Maybe with the `can()` helper we could use this syntax:
```ts
can.view(ontology.objects())
can.view(ontology.objects().except([AuditLog]))

can.apply(actions().except([dangerousAction]))
can.apply([sendQuote, cancelQuote])

can.start(workflows().except([sendMoney]))
can.start([followUpQuote])
```

We could have `ontology`, `actions`, `workflows` under the core security:

```ts
// packages/core/src/security/scopes.ts
export const ontology = {
  objects: () => objectTypeScope({ include: "all", except: [] }),
}

export const actions = () => actionScope({ include: "all", except: [] })

export const workflows = () => workflowScope({ include: "all", except: [] })
```

```ts
import { actions, can, defineRole, ontology } from "@sixb/core"

export const admin = defineRole("admin", {
  grantedTo: [admins],
  grants: [
    can.view(ontology.objects().except([AuditLog])),
    can.apply(actions().except([dangerousAction])),
  ],
})
```

On the ability to view events, maybe we don't have `viewAllEvents()` and the user is just able to view events based on its authorizations on objects, workflows, actions.
Like if I have the permission `can.view(ontology.objects().except([AuditLog]))`, I should be able to see every objects events except those from `AuditLog`. Same with workflows and actions, I can see the events of the `requests`, `runs` lifecycle for those I have access. What do you think ?

packages/core/src/authorization/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  readonly canApplyAllActions: boolean
+  /** Action ids this principal may request. */
+  readonly applicableActionIds: ReadonlySet<string>
+  /** Whether this principal may run every workflow. */
+  readonly canRunAllWorkflows: boolean
+  /** Workflow ids this principal may run. */
+  readonly runnableWorkflowIds: ReadonlySet<string>
+  /** Whether this principal may read the domain event stream. */
+  readonly canViewEvents: boolean
+}
```
quentinnippert, line 26: Based on my comment on the `can` syntax, I would do:

```ts
interface GrantIndex {
  objectTypes: {
    view: ReadonlySet<string>
  }

  actions: {
    apply: ReadonlySet<string>
  }

  workflows: {
    start: ReadonlySet<string>
  }
}
```
I would avoid wildcard booleans like `canViewAllObjects` in the resolved index. Broad grants should be expanded at startup after applying `except(...)`, so runtime checks stay simple `set.has(id)` checks.

packages/core/src/objects/sdk/object-set.ts
```diff
@@ -71,12 +68,14 @@ export function createObjectSet<
...
     storage,
     queues,
+    authorization,
     objectType,
     primaryPropertyId,
   }
 
   const objectSet = {
     get: async (id: string) => {
+      assertCanView(resolvedCtx, objectType.id)
```
quentinnippert: Could be good to have a security engine that we can reuse like this that would centralize everything:
```
const decision = security.authorize({
  principal,
  action: "object.view",
  resource: objectRef(Quote, "q1"),
  context,
})
```

Same for query, actions, workflows...
That makes me think about `query`, wondering how that lives with objects authorizations and how we validate the query is authorized.

---

Codex came up with this:

I would see it like this: `query` should not be a simple `authorize(...): allow/deny`. A query is a **read program**, so authorization needs to do two things:

```txt
1. validate: does the query touch authorized resources?
2. rewrite/filter: how do we limit the results to visible objects?
```

I would propose a dedicated API:

```ts
const authorized = security.authorizeObjectQuery({
  principal,
  query,
  operation: "object.query",
})
```

which would return something more like:

```ts
{
  decision: "allow",
  query: authorizedQuery,
  requirements: [
    "object.view:Quote",
    "object.view:Contact",
    "property.read:Quote.margin",
  ],
}
```

Not just:

```ts
true
```

**Flow**

```ts
const normalized = normalizeQuery(query)

const requirements = analyzeQuery(normalized)
// object types touched, traversed links, projected properties,
// predicates, sorts, facets, text/vector fields...

const decision = security.authorize({
  principal,
  action: "object.query",
  requirements,
})

const securedQuery = security.rewriteQuery({
  query: normalized,
  principal,
  policies,
})

const rows = await executeQuery(securedQuery)

return redactProperties(rows, principal)
```

quentinnippert (commented):
Great work laying out all this bro, some comments

packages/core/src/security/types.ts
demattosanthony: yeah makes sense. much better DX. pushed the updates

packages/core/src/authorization/types.ts
demattosanthony: pushed some updates for this as well

packages/core/src/objects/sdk/object-set.ts
demattosanthony: Agreed on a central engine. I think it belongs in the authorization layer rather than security, so that's where I put it. I added an `evaluate` method that returns the decision, plus `isAllowed`/`assertAuthorized` derived from it, that way we get both "give me the decision" and "just throw" without duplicating the checks:

```ts
evaluate(ctx, request): { allowed, requirements, missing }
isAllowed(ctx, request): boolean              // = evaluate(...).allowed
assertAuthorized(runtime, request): void      // throws unless allowed
```

Everything routes through it now (reads, actions, workflows, query, events, server routes) via a typed `request`. I could see this later powering something like `sixb.as(ctx).can(...)`, but there's no use case yet so I left that out.

On the query, I dug into it and it's basically already doing what you described. `validateObjectQuery` walks the query and collects every object type it touches (including traversal hops), and we authorize all of them at plan time, before any storage call:

```ts
assertAuthorized({ authorization }, { kind: "object.query", touchedObjectTypeIds })
```

So the "does the query touch authorized resources" step is covered, I left that flow as is.

The rewrite/filter + redaction part I left out for now: grants are type-level only (no per-row or per-property), so if you can view a type you see all its rows and columns, and if you can't the query is just denied. Nothing to rewrite or redact yet. Easy to slot in later if we add row/column grants.

quentinnippert:
It's very clean I love it. Added some small updates, maybe check last commit and we merge if you're all good with the changes ?

quentinnippert (approved):

demattosanthony:
> It's very clean I love it. Added some small updates, maybe check last commit and we merge if you're all good with the changes ?

looks great. i love it

PR #77, opened 2026-06-17
atlas: add object query builder

demattosanthony:
## Summary
- expose ontology query/search metadata through object type responses and regenerated client types
- add a server-backed Atlas query builder for selected object types with text search, filters, match-all/match-any, sort, facets, and load-more pagination
- split object query UI/model into focused Atlas components and helpers, and remove extra per-object detail queries from the workbench

## Validation
- bun run generate:client
- bun test packages/server/tests/openapi-docs.test.ts
- bun run check
- bun run typecheck
- bun run build && bun run test

## Screenshots

Screenshot 2026-06-16 at 10 50 18 PM
Screenshot 2026-06-16 at 10 50 24 PM
Screenshot 2026-06-16 at 10 50 46 PM

### DIFF
--- a/packages/atlas/src/components/objects/ObjectFilterPopover.tsx
+++ b/packages/atlas/src/components/objects/ObjectFilterPopover.tsx
@@ -1,0 +1,315 @@
+import type { ObjectQueryFacetResult } from "@sixb/client"
+import {
+  Button,
+  Input,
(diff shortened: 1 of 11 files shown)


quentinnippert:
it's awesome, tested on my sixb, it works well

PR #87, opened 2026-06-21
workers: idle instead of exiting when a type has no definitions

demattosanthony:
A dedicated worker process (`sixb worker <type>`) started for a type with **no** registered definitions threw in its constructor → the process exited non-zero → the supervisor restarted it forever (we saw 94 restarts).

```
[SixbSyncWorker] No sync definitions are registered.   → exit 1 → restart loop
```

Now an empty worker **idles** (waits for shutdown) instead of crashing — matching what `action-worker` already did.

### Change
- `QueueWorker` gains an `idle` flag; when set, `run()` waits for the abort signal instead of polling the queue (single source of truth).
- `sync` / `pipeline` / `projection` / `workflow` workers compute `idle = definitions.length === 0`, log once, and idle instead of throwing. `action-worker` is refactored to use the shared flag (drops its hand-rolled `run()` override).
- Run-storage requirement checks (`storage.syncRuns`, etc.) now fire only when definitions exist — an empty worker no longer needs run storage to idle.

| Worker | empty before | empty after |
|---|---|---|
| sync / pipeline / projection / workflow | throw → exit 1 | idle |
| action | idle (already) | idle (shared flag) |

The cohosted path (`sixb dev`) already guarded each worker, so only dedicated worker processes were affected.

### Tests
- Per-worker "idles without definitions" unit tests + a base `QueueWorker` idle test.
- The CLI provider-cleanup tests asserted the old crash; since "no definitions" is no longer a failure (and would hang `spawnSync` on an idling worker), they were repointed to a genuine post-load failure via a new `worker-missing-run-storage` fixture (registers a sync but omits `storage.syncRuns`).

Note: `rules-worker` still throws on empty but is unaffected — it's not a dedicated worker type and is guarded by the runtime.

### DIFF
--- a/packages/action-worker/src/worker.ts
+++ b/packages/action-worker/src/worker.ts
@@ -34,9 +34,10 @@ export interface ActionWorkerOptions {
 export class ActionWorker extends QueueWorker<ActionRunRequestedQueueJob> {
   private readonly context: ActionWorkerContext | null
   private readonly sixb: ActionWorkerSixb
-  private readonly idleWithoutDefinitions: boolean
 
   constructor(sixb: ActionWorkerSixb, options: ActionWorkerOptions = {}) {
+    const idle = sixb.getActionDefinitions().length === 0
+
     super({
       projectId: sixb.id,
       queue: sixb.queues.actions,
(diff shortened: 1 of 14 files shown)


quentinnippert:
do you think we could model the empty case as its own worker instead of an idle flag on each one?

An `IdleWorker` would extends `Worker` that just waits for abort, returned by `createWorkerForType` when the count is 0. The real workers keep a non-null context and we can drop the `requireContext()` guards

PR #113, opened 2026-06-24
agents: add message model + persistence contract (in-memory)

quentinnippert:
# Agents — message model + durable persistence

SDK-independent persistence for agent conversations: a clean message model + adapters, a 3-table
storage contract, and **three backends behind it — in-memory, SQLite, Postgres** — all proven by one
shared contract suite.

## ✨ At a glance

```ts
import { fromAiSdk, toModelMessages, type SixbMessage } from "@sixb/core"

const message: SixbMessage = fromAiSdk(uiMessage)   // SDK shape → our own model (total: throws, never drops)
const modelMessages = toModelMessages(history)      // replay back into a model

// Same API on any backend — InMemoryStorage / SqliteStorage / PostgresStorage
const run = await storage.agents.runs.reserve({ id, projectId, threadId, agentId, triggerMessageId, lease })
await storage.transaction(async (tx) => {
  await tx.agents.messages.append({ /* role + parts (+ metadata) */ })
  await tx.agents.runs.finish({ projectId, id: run.id, leaseId, status: "succeeded", usage })
})
```

## 📦 What's inside

**`SixbMessage`** — a `role` + structured `parts` + optional `metadata`. No SDK wrapper, no `raw`
blob; the row owns identity, a `contentVersion` column versions the parts shape.

| part | carries |
|---|---|
| `text` / `reasoning` | text (+ provider metadata, e.g. a reasoning signature) |
| `step-start` | a loop step boundary |
| `tool-call` | tool name · input · terminal output / error |

**Adapters** — `fromAiSdk` / `toUiMessage` / `toModelMessages`, all on our own types (**core never
imports `ai`**). `fromAiSdk` is **total**: it throws rather than silently drop an unmodeled part, a
transient/streaming state, a bad role, or a non-JSON payload.

**Three tables, tiered store**

```
agents
├── threads    create · getById · list          — one conversation · single-flight anchor (active_run_id) · stats
├── runs       reserve · renewLease · reclaim ·  — one loop execution · lease (id + expiry) · attempt · usage
│              finish · getById · list
└── messages   append · getById · list           — append-only · per-thread seq · role · parts (JSON) · metadata
```

- **single-flight** — a thread has at most one active run (`reserve` claims it, `finish` releases it).
- **lease** — `renewLease` heartbeats, `reclaim` takes over an expired lease (crash recovery).
- **seq** — assigned per thread on append; messages are written once a run finalizes.

**Three backends, one spec** — `InMemoryAgentStorage` (`Map` + `structuredClone` + snapshot/restore),
plus **SQLite** and **Postgres** adapters (`storage/{sqlite,pg}/src/agents/`). All three are replayed
against the same `runAgentStorageContractSuite`, so **in-memory ≡ sqlite ≡ pg**. Atomicity reuses the
existing `Storage.transaction()` (`FOR UPDATE` on pg, `BEGIN IMMEDIATE` on sqlite) — no god-method.

## 🧭 Key choices

- **Own the schema** — no dependency on `ai`, no `raw_sdk_message` blob.
- **A message is a document** — `parts` is stored as canonical JSON (lossless, exactly what the
  adapters replay). No projection table to keep in sync; part-level search can use a JSONB index later.
- **Flatten what's fixed** — `owner_principal` → `type`/`id` columns, `usage` → 5 int columns. The core
  records keep their object shape; only the SQL row-mappers flatten. No JSON blobs for scalar data.
- **No foreign keys** — referential integrity stays in app code (the contract already requires it),
  keeping sqlite and pg structurally identical; dialects differ only in column types (`TIMESTAMPTZ`/
  `JSONB` vs ISO `TEXT`/JSON `TEXT`).
- **Total adapter** — unmodeled input throws, signalling exactly when to extend the part union.

## 🚫 Out of scope (later slices)

Worker loop · stream broker + WebSocket · routes + client · run-token (agent principal) · UI.

## ✅ Verification

| Check | Result |
|---|---|
| Core — typecheck + tests | ✅ 878 pass |
| SQLite — typecheck + tests | ✅ 176 pass (contract replay + bundle rollback + on-disk migration) |
| **Postgres — e2e on real Postgres** | ✅ 22 pass (contract replay + bundle rollback + migration) |
| Repo-wide typecheck · build · generated client | ✅ no drift (no API surface) |


packages/core/src/storage/agents/types.ts
```diff
@@ -0,0 +1,238 @@
...
+export interface AgentRunRecord {
+  readonly id: string
+  readonly projectId: string
+  readonly threadId: string
+  readonly agentId: string
+  readonly triggerMessageId: string
+  readonly status: AgentRunStatus
+  readonly modelId?: string
+  /** AI SDK `FinishReason`, stored as an opaque string (core does not depend on `ai`). */
+  readonly finishReason?: string
```
demattosanthony: even though we don't depend on ai sdk maybe we can still type the finish reason as our own internal enum

packages/core/src/agents/message.ts
```diff
@@ -0,0 +1,79 @@
...
+export type SixbMessagePartType = SixbMessagePart["type"]
+export type SixbToolCallState = SixbToolCallPart["state"]
+
+/** The logical agent message the adapters speak and the store persists (as columns). */
+export interface SixbMessage {
+  readonly role: SixbMessageRole
+  readonly parts: readonly SixbMessagePart[]
+  /** Message-level metadata (mirrors `UIMessage.metadata`). */
+  readonly metadata?: JsonValue
+}
```
demattosanthony: should we remove the `Sixb` prefix from all the types here? if we want a prefix maybe just `AgentMessage` or something like that? 

Same with the adapters.ts file

demattosanthony (commented):
looks great. love the separation and layers of threads runs and messages

demattosanthony (approved):

PR #124, opened 2026-06-27
atlas: rebuild the dataset viewer as a full-height data grid

demattosanthony:
## Problem

The dataset detail page boxed its data into a doubly-constrained window: the route sat inside the app's `max-w-7xl` shell, the page added its own `max-w-6xl`, a 320px metadata sidebar took the rest, and the row table was capped at `max-h-[520px]`. The data — the point of the page — was the smallest thing on screen, with no way to sort, resize columns, filter, or inspect long/JSON values.

## Change

Move the route out of the constrained shell (like the pipeline/workflow detail pages) and rebuild the page around a reusable grid. Four bands, each with one job:

## Scope

UI only — `packages/atlas` (dataset page + pipeline preview) and one new `packages/ui` component. No server, route, schema, or generated-client changes.

Screenshot 2026-06-26 at 11 06 29 PM

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -248,11 +248,11 @@
       "name": "@sixb/agent-worker",
       "version": "0.1.0",
       "dependencies": {
-        "@ai-sdk/provider": "^3.0.10",
         "@sixb/core": "workspace:*",
         "ai": "^6.0.0",
       },
       "devDependencies": {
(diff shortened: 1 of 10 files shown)


quentinnippert (approved):
love it

PR #224, opened 2026-07-17
core: add staged object upsert

demattosanthony:
## Summary

Adds a staged object upsert to the action `.edits()` facade:

```ts
objects(Contact).upsert({
  id,
  ...properties,
})
```

This replaces the current read-before-write pattern:

```ts
const existing = await read.objects(Contact).byId(id).get()

if (existing) {
  objects(Contact).byId(id).update(properties)
} else {
  objects(Contact).create({ id, ...properties })
}
```

## Semantics

- requires the object type's actual primary property
- creates when the object is absent
- merges supplied properties when it exists
- preserves omitted properties on update
- validates all required properties when the operation resolves to create
- returns an edit handle for follow-up updates, links, or deletion
- commits with the action's other edits in the existing serializable transaction
- emits `object.created` or `object.updated` for the net change; unchanged upserts are no-ops

The create/update decision is represented as `object.upsert` in the transient EditBatch and resolved by the planner inside the transaction. Storage providers continue receiving the existing concrete create/update commit plans, so no migrations or provider primitives are needed.

## Coverage

Includes type-level primary/property checks, raw operation validation, partial update and required-on-create behavior, ordered in-batch edits, delete/recreate behavior, no-op folding, mutation events, worker integration, concurrency, and SQLite commit-plan regression coverage.

## Verification

- `bun test packages/core/tests/edits.test.ts packages/action-worker/tests/run-action-job.test.ts`
- `bun test storage/sqlite/tests/sqlite-edit-commit.test.ts`
- `bun run typecheck`
- `bun run build`
- `bun run check`


quentinnippert:
love that

quentinnippert (approved):

PR #229, opened 2026-07-18
blob-s3: bound streaming upload memory

quentinnippert:
## Problem

`Bun.S3Client` can retain native allocations during long sequences of streamed uploads on Bun
1.3.14/Linux.

```text
SFTP stream → Bun multipart writer → native RSS keeps growing → worker killed
```

Applications were forced to bypass `BlobStorage.put()` to keep large syncs stable. The provider
also mixed Bun's client, hand-written SigV4 requests, and the AWS SDK, which gave S3 operations
different credential, retry, and error behavior.

## Fix

`@sixb/blob-s3` now uses one long-lived AWS SDK client for every S3 operation:

- fixed-size, replayable upload parts;
- bounded multipart concurrency;
- AWS SDK retries and per-part `Content-MD5`;
- multipart abort and source cancellation on failure;
- streamed `GetObject` and ranged reads without buffering;
- `HeadObject`, `CopyObject`, and `DeleteObject` for metadata and cleanup;
- official AWS request presigning for checksum-protected direct uploads;
- one credential chain and endpoint/path-style configuration across the provider.

The Bun S3 client, lazy upload-only client, custom retry loop, and hand-written SigV4 signer have
been removed.

| Option | Default | Effect |
| --- | ---: | --- |
| `putPartSizeBytes` | 8 MiB | Buffer size per part |
| `putConcurrency` | 2 | Maximum concurrent part uploads |
| `putRetries` | 3 | Retries for replayable AWS SDK requests |

Approximate provider-owned upload-buffer memory remains bounded by:

```text
putPartSizeBytes × putConcurrency
```

No connector or application workaround is required.

## Validation

- multipart ordering and bounded concurrency;
- real AWS SDK retry behavior against transient 503 responses;
- Web-stream and range-read behavior;
- CopyObject source encoding;
- signed checksum, content-type, ACL, and session-token headers;
- cancellation, abort, and staging cleanup;
- expected-size and backend-verified checksum validation;
- MinIO small, empty, multipart, range, stat, missing, and direct uploads;
- package build and typecheck;
- repository-wide typecheck and lint;
- publish smoke test for all 44 publishable packages.


quentinnippert:
@demattosanthony updated

PR #246, opened 2026-07-22
workflows: add structured agent steps

### LINKED ISSUE #182: Idea: add structured read-only agent steps to workflows
# Proposal: add structured read-only agent steps to workflows

## Problem

Projects currently put AI SDK calls inside generic workflow steps:

```ts
defineWorkflowStep("resolve-transcript")
  .input({ transcript: ref(Transcript) })
  .output({ projectId: "string", confidence: "double" })
  .run(async ({ input, sixb }) => {
    const candidates = await loadCandidates(sixb)
    return generateStructuredResolution(input.transcript, candidates)
  })
```

This is fine for a bounded, one-shot inference, but it does not compose naturally when the model
needs to loop, inspect Sixb data, and adapt its search from intermediate results.

It also mixes four concerns in project code:

```txt
data retrieval + agent loop + output validation + business effects
```

Sixb already has `defineAgent(...)`, scoped agent identities, sandboxed tools, and typed workflow
dataflow. Workflows need a first-class way to invoke those agents and pass a structured result to
the next node.

## Proposal

Add a read-only `defineAgentStep(...)` primitive that references an existing agent definition.

```ts
// agents/transcript-resolver.ts
export const transcriptResolver = defineAgent("transcript-resolver", {
  name: "Transcript Resolver",
  model: gateway("anthropic/claude-sonnet-4.5"),
  instructions: [
    "Match meeting transcripts to active projects.",
    "Ground every answer in Sixb data.",
    "Return no match when evidence is insufficient.",
  ].join("\n"),
  groups: [operations],
  loop: { stopWhen: { maxSteps: 8 } },
})
```

```ts
// workflows/resolve-transcript.ts
const resolveTranscript = defineAgentStep("resolve-transcript", transcriptResolver)
  .input({
    transcript: ref(Transcript),
  })
  .output({
    projectId: "string",
    confidence: "double",
    reason: "string",
  })
  .prompt(({ input }) =>
    [
      `Resolve transcript '${input.transcript.primaryId}'.`,
      "Search active projects and their related clients.",
      "Return the best match and the evidence behind it.",
    ].join("\n")
  )
```

The agent output participates in workflow dataflow like any other typed node:

```ts
export const resolveTranscriptWorkflow = defineWorkflow("resolve-transcript")
  .input({ transcript: ref(Transcript) })
  .then(resolveTranscript)
  .then(applyResolution, ({ input, steps }) => ({
    transcript: input.transcript,
    projectId: steps.resolveTranscript.projectId,
    confidence: steps.resolveTranscript.confidence,
    reason: steps.resolveTranscript.reason,
  }))
```

`defineAgent` owns the reusable actor: model, instructions, identity, groups, and loop defaults.
`defineAgentStep` owns one workflow task: node id, input, prompt, and structured output.

The output contract belongs to the step, not to the agent, because one agent may perform several
tasks with different outputs.

## Read-Only Boundary

Agent steps are read-only in V1.

```txt
effective access = agent grants ∩ read-only agent gateway routes
```

The agent may inspect the ontology, objects, links, datasets, telemetry, and other explicitly
allowlisted read surfaces. It cannot request actions or mutate objects, links, or telemetry.

Effects remain explicit workflow nodes:

```txt
agent research
  -> structured proposal
  -> deterministic validation / intervention
  -> action node
```

This keeps untrusted data and prompt injection away from mutation authority. Structured output
validates shape, but the following deterministic node remains responsible for business policy.

The step reuses the agent's existing service-account identity, sandbox, restricted network, and
run-scoped gateway capability. It must not receive the workflow worker's privileged `sixb`
runtime.

For manual runs, the caller must be allowed to run both the workflow and every referenced agent.
Scheduled runs execute as the system principal. Every workflow run records its
`requestedByPrincipal` so the delegation remains auditable.

## Runtime Shape

An agent step is a durable child execution, not an AI SDK call inside the workflow worker.

```txt
workflow worker
  -> start agent node
  -> enqueue child agent task run
  -> workflow waits

agent worker
  -> resolve agent + workflow node definitions
  -> create sandbox + read-only gateway capability
  -> run tool loop
  -> validate structured output
  -> persist trace, usage, and result
  -> emit completion

workflow worker
  -> resume idempotently
  -> store node output
  -> continue dataflow
```

The child run id is deterministic from the workflow run and node index. Redelivery must reuse the
same completed result instead of running the model twice.

**Workflow task runs are headless and start with fresh context. They do not create a fake conversation thread or share memory across workflow runs.**

The workflow node stores business input/output plus a reference to the child agent run. The agent
run remains the source of truth for model id, usage, tool calls, finish reason, and errors.

## Structured Output

The Sixb `.output(...)` declaration is the only public schema.

```txt
Sixb output schema
  -> internal AI SDK structured-output adapter
  -> model result
  -> Sixb runtime validation
  -> workflow node output
```

Projects should not maintain both a Zod schema for the AI SDK and a Sixb schema for the workflow.
The agent worker converts the Sixb contract internally and validates the final value again before
the workflow resumes.

An invalid or missing structured output fails the agent node. Partial text and tool traces remain
available for inspection but never flow to the next node.

## Why Not Reuse Existing Surfaces Directly?

- `.then(agent, { ... })` overloads workflow composition with prompt, schema, and node identity
  concerns, and becomes awkward when the same agent is used more than once.
- Adding `.output(...)` to `defineAgent` couples a reusable actor to one task contract.
- Posting a synthetic chat message creates thread ownership and memory semantics that a workflow
  task does not need.
- Direct AI SDK calls remain appropriate for bounded one-shot inference; this proposal does not
  force every model call through an agent loop.

## Done Looks Like

- `defineAgentStep(id, agent)` declares typed input, prompt, and output.
- Workflows accept agent steps as nodes with `type: "agent"`.
- Startup validation rejects unknown agents and unauthorized workflow-to-agent delegation.
- Agent steps expose only allowlisted read-only Sixb routes.
- The agent worker executes headless task runs with the existing identity, sandbox, and loop.
- Structured output uses the Sixb workflow schema as its public source of truth.
- Workflow runs wait and resume from durable agent completion without holding a worker lease.
- Child run creation, completion, cancellation, and resume are idempotent.
- Workflow node APIs expose the child agent run id; agent run APIs retain trace and usage.
- Tests cover type inference, access denial, invalid output, step limits, cancellation, redelivery,
  and workflow continuation.

## Non-Goals For V1

- Agent-initiated actions or any other mutation tool.
- Tool approval flows.
- Custom project tools.
- Shared memory or reuse of conversation threads.
- Agent batches, `foreach`, sub-agents, or agent-to-agent delegation.
- Replacing simple one-shot structured AI SDK calls.

## Still Unclear

- Whether headless task executions should extend `agent_runs` with a `mode` and parent reference, or
  use a separate storage record while sharing the same execution engine.

quentinnippert:
Closes #182.

## Context

Sixb already had the two ends of the problem:

| Primitive | Responsibility |
| --- | --- |
| `defineAgent` | Reusable actor: model, instructions, identity, grants, tools and loop limits |
| `defineWorkflow` | Durable orchestration: typed dataflow, waiting, resume and deterministic effects |

What was missing was a typed, durable bridge between them.

Projects had to hide agentic work inside a generic workflow step:

```text
workflow step
  └─ project-owned AI SDK call
       ├─ agent loop
       ├─ Sixb data access
       └─ structured-output validation
```

That mixed workflow orchestration, agent execution and output validation in project code. This PR introduces a first-class agent node that composes with the existing primitives.

## Authoring experience

The agent remains reusable. The workflow task owns its input, prompt and output contract:

```ts
const resolveTranscript = defineAgentStep("resolve-transcript", transcriptResolver)
  .input({
    transcript: ref(Transcript),
  })
  .output({
    projectId: "string",
    confidence: "double",
    reason: "string",
  })
  .prompt(({ input }) =>
    [
      `Resolve transcript '${input.transcript.primaryId}'.`,
      "Search active projects and their related clients.",
      "Return the best match and the evidence behind it.",
    ].join("\n")
  )
```

It participates in workflow dataflow like any other typed node:

```ts
defineWorkflow("resolve-transcript")
  .input({ transcript: ref(Transcript) })
  .then(resolveTranscript)
  .then(applyResolution, ({ input, steps }) => ({
    transcript: input.transcript,
    projectId: steps.resolveTranscript.projectId,
    confidence: steps.resolveTranscript.confidence,
    reason: steps.resolveTranscript.reason,
  }))
```

| Primitive | Owns |
| --- | --- |
| `defineAgent` | How the actor reasons and what it may access |
| `defineAgentStep` | One task's input, prompt and structured output |
| `defineWorkflow` | When the task runs and where its result flows next |

## Durable execution

```mermaid
sequenceDiagram
  participant W as Workflow worker
  participant S as Workflow storage
  participant Q as Agent queue
  participant A as Agent worker

  W->>S: Create agent node execution
  W->>S: Mark node + workflow waiting
  W->>Q: Enqueue deterministic node job
  Note over W,S: Workflow worker releases its delivery
  Q->>A: Deliver agent node
  A->>A: Sandbox + Sixb tools + agent loop
  A->>A: Produce and validate structured output
  A->>S: Persist output, trace, usage and finish reason
  A-->>W: Enqueue workflow resume
  W->>S: Reclaim workflow and complete resume
  W->>W: Continue typed dataflow
```

The workflow does not hold a worker lease while the agent runs. The agent execution is headless, starts with fresh context and does not create a synthetic conversation thread.

## Output contract

The Sixb `.output(...)` declaration remains the only public schema:

```text
Sixb step output
      │
      ├──> JSON Schema adapter ──> AI SDK structured generation
      │
      └──> Sixb runtime validation
                                │
                                ▼
                    workflow_node_runs.output
```

Invalid or missing structured output fails the node and never reaches the next workflow step.

## Persistence model

```text
workflow_runs
└── workflow_node_runs                  business input/output + node status
    └── workflow_agent_node_runs        prompt + agent execution metadata
        ├── model / finish reason
        ├── usage / trace / diagnostics
        └── attempt / execution lease / error
```

`workflow_agent_node_runs` has a one-to-one relationship with its workflow node run. The initial PostgreSQL and SQLite schemas are updated directly; this PR does not add a new migration.

## Guarantees

| Concern | Behavior |
| --- | --- |
| Typing | Input is inferred in `prompt`; output is inferred in downstream `steps` |
| Redelivery | Deterministic node identity, reclaim attempts and execution-token fencing |
| Completion | Agent metadata and workflow node output are finalized transactionally |
| Cancellation | Workflow, node and agent execution transition coherently |
| Observability | Prompt, trace, usage, diagnostics, model and errors are exposed through the workflow API and Atlas |
| Storage parity | In-memory, SQLite and PostgreSQL implement the same contract |

## V1 decisions

- Agent steps use the same identity, sandbox, tools, gateway and loop configuration as conversational `defineAgent` executions.
- No workflow-specific read-only restriction is imposed in V1; effective access is governed by the referenced agent's grants and the existing agent gateway.
- The output contract belongs to `defineAgentStep`, not `defineAgent`, so one agent can perform tasks with different result shapes.
- No extra execution/groups configuration is added to the step.
- No conversation memory, batches, sub-agents or agent-to-agent delegation are introduced.

## Main implementation surfaces

| Area | Change |
| --- | --- |
| Core | `defineAgentStep`, type inference, validation, snapshots and JSON Schema conversion |
| Workflow worker | Park/resume orchestration, recovery, queue ownership and agent-node executor |
| Agent worker | Headless structured execution using the existing agent environment |
| Storage | Durable workflow-agent node lifecycle across all providers |
| Server / Atlas | Diagnostics, trace, usage and cancellation visibility |
| Client | Regenerated OpenAPI and typed client contracts |

## Validation

- `bun run generate:client`
- `bun run check`
- `bun run typecheck`
- `bun run test` — **2,757 passed**, 2 skipped
- `bun --filter @sixb/pg test:e2e` — **188 passed**


packages/core/src/runtime/scoped.ts
```diff
@@ -218,10 +218,23 @@ export function createScopedSixb<TOntologySources extends readonly OntologySourc
...
+          agentId: node.agentStep.agent.id,
+        })
+    )
+  const workflows = {
+    list: () => deps.workflows.list().filter(canRunWorkflow),
+    getById: (workflowId: string) => {
+      const workflow = deps.workflows.getById(workflowId)
+      return workflow && canRunWorkflow(workflow) ? workflow : null
+    },
+  }
```
demattosanthony: should the scope remain just on the workflow boundary? Feels like coupling the workflow access with the agent access could be tricky

not sure

demattosanthony (approved):

packages/core/src/runtime/scoped.ts
quentinnippert: pushed some updates

Issue #237, opened 2026-07-18
Phase 6: Delete legacy paths and ship the clean break


demattosanthony:
# Phase 6: Delete legacy paths and ship the clean break

## Outcome

Remove every superseded mutation path, finalize rewritten initial schemas and public contracts, prove
there is one write architecture, and ship without a compatibility runtime.

## Release invariants

- Action, runtime object API, projection replacement, and telemetry use `OntologyMaterializer`.
- Every provider passes one conformance suite.
- No mutation event is used as a storage command.
- No old provider mutation method remains callable.
- No V1/V2 EditBatch branch remains.
- No caller-managed projection staging API remains.
- No CDC cursor/order/tombstone/delta contract is included.
- No intermediate feature flag or fallback exists.

## Delete

Core/provider mutation paths:

- event-before-state object upsert/link upsert/link remove;
- event-command telemetry append/write;
- applied-event dedupe/provenance used for storage mutation;
- Action-specific physical planner/application/provider methods;
- provider object/link/timeseries write methods outside `OntologyStorage`;
- duplicated mutation-event/property diff builders;
- old transaction helpers needed only by removed write paths.

Action compatibility:

- old EditBatch version/type normalization branches;
- reconstructed Action commit `EventDraft[]`;
- separate Action physical diff persistence;
- Action worker mutation publication.

Projection compatibility:

- projection calls to `sixb.objects(...).upsert/link/unlink`;
- projection public CRUD batching helpers;
- exposed staged snapshot/checkpoint/discard contracts;
- dormant CDC/change-stream code or schema columns.

Operational compatibility:

- migration manifests/import/bootstrap/rebuild services;
- ontology migration CLI proposals;
- legacy data inspection endpoints;
- compatibility/shadow/dual-write modes.

## Keep and narrow

- `packages/core/src/edits/` remains typed authoring/normalization only.
- Runtime object SDK/service/leaf files authorize/normalize, invoke one materializer adapter, and map
  results.
- Projection worker maps immutable rows to lazy source entries and calls one replacement API.
- `EventsRuntime.publishEnvelopes` remains internal to the outbox dispatcher.
- Public custom events remain supported but reserved mutation event types are rejected.

## Static architecture gates

Add script/test checks that fail on:

- `events.append(` in object/link/telemetry/Action/projection mutation modules;
- object/link/timeseries write methods on read-storage interfaces;
- projection-worker imports of runtime object mutation modules;
- old provider mutation method names;
[...]

Issue #236, opened 2026-07-18
Phase 5: Integrate hosting, rules, effects, observability, and public contracts


demattosanthony:
# Phase 5: Host post-commit delivery, rules reconciliation, maintenance, and operations

## Outcome

Complete the post-commit runtime around the Materializer without adding CDC, compatibility
migrations, or a generic observability framework.

```text
atomic ontology commit
  -> durable ontology_outbox row
  -> immediate best-effort drain
  -> periodic catch-up and retention maintenance
```

The database remains authoritative. Broker availability affects delivery latency only.

## Outbox delivery

The Materializer atomically writes ontology state, `ontology_commits`, and stable event envelopes in
`ontology_outbox`.

After a successful commit:

- `notify()` starts one coalesced, fire-and-forget `drain()`;
- the mutation never waits for the broker;
- there is no polling loop in standalone workers;
- failed rows are rescheduled with the existing exponential backoff and jitter;
- leases and stable envelope IDs preserve multi-publisher safety and at-least-once delivery.

Batch publication failure must not let one invalid envelope block valid siblings:

```text
batch fails
  -> bounded isolation of failing envelopes
  -> mark successful subsets published
  -> reschedule only failures
```

Isolation work is bounded per pass. Failed rows remain durable and retryable; Phase 5 adds no silent
drop or quarantine.

## `OntologyMaintenance`

One lifecycle service owns recovery and retention:

```text
OntologyMaintenance
  |- outbox catch-up
  |- published outbox cleanup
  `- terminal source cleanup
```

Defaults:

| Setting | Default |
|---|---:|
| catch-up / cleanup interval | 60 seconds |
| published outbox retention | 24 hours |
| terminal source retention | 24 hours |
| cleanup limit per domain/pass | 1,000 |

All values are configurable. Passes do not overlap.

Hosting:

- `SixbServer.start()` starts maintenance before serving; `stop()` releases it before providers;
- `sixb dev` gets the same single service through its server;
- standalone workers perform only immediate post-commit drains;
- an embedded runtime without a server explicitly calls `sixb.startOntologyMaintenance()`;
- the `Sixb` constructor starts no timer or polling loop;
- Phase 5 adds no dedicated outbox worker.

Catch-up claims only due unpublished rows through the indexed, bounded lease contract. It runs once
at startup, then every interval. A crash between commit and immediate drain is therefore recoverable.

Retention rules:

- purge only outbox rows with `published_at` older than retention;
[...]

Issue #235, opened 2026-07-18
Phase 4: Route projection replacement and telemetry through the materializer


demattosanthony:
# Phase 4: Route projections through the ontology materializer

## Outcome

Route object/link source replacement and telemetry projections through the internal
`OntologyMutationRuntime` and authoritative ontology materializer.

```text
dataset.version.committed
  -> deterministic projection job/run
  -> QueueDelivery claim
  -> fenced projection execution
  -> immutable dataset read
  -> OntologyMutationRuntime
  -> Materializer
  -> source/effective state + ontology commit + outbox
```

The implementation must provide:

- one deterministic logical run for one full projection/dataset semantic identity;
- lazy, bounded-memory object/link replacement;
- resumable, fixed-boundary telemetry ingestion;
- token-fenced writes across queue redeliveries;
- automatic post-commit outbox wake-up;
- physical run progress separated from semantic commit results.

## Scope

Included:

- deterministic dispatch identity and queue payloads;
- `ProjectionWorker` migration from public CRUD to the internal mutation runtime;
- object/link replacement source mapping;
- telemetry projection batching/checkpoint completion;
- stable immutable-version row iteration contract;
- projection-run progress cleanup;
- PostgreSQL, SQLite, in-memory, local-lake, and DuckLake provider changes;
- server/OpenAPI/client/UI updates required by the run progress contract.

Already completed in Phase 3 and not reimplemented here:

- public runtime telemetry APIs already route through `OntologyMutationRuntime`;
- public authoring restrictions for object/link/telemetry facts;
- Action edit materialization.

Explicitly deferred:

- CDC, deltas, tombstones, cursors, row IDs, and change streams;
- a framework/materializer build revision;
- writeback read-dependency persistence;
- terminal source cleanup and full operational observability (`#236`).

Schema policy for this phase:

- edit `001-initial-schema.sql` directly for PostgreSQL and SQLite;
- do not add a new migration;
- existing databases must be recreated; compatibility with an already-applied `001` is not
  provided by this phase.

## Ownership boundaries

| Component | Responsibility |
|---|---|
| Dataset event | Publish the exact immutable dataset version metadata |
| `ProjectionRegistry` | Resolve definitions, ownership, and semantic revisions |
| Orchestrator | Build the deterministic job; no ontology recomputation |
| `QueueDelivery` | Lease renewal, loss detection, retry/fail/complete settlement |
[...]

Issue #234, opened 2026-07-18
Phase 3: Route Actions and the runtime object API through durable edits


demattosanthony:
# Phase 3: Route Actions and the runtime object API through durable edits

## Outcome

Replace the current Action planner and event-first runtime object/link writes with thin producers of
`OntologyEditCommit`. Keep runtime object API signatures. Replace the EditBatch contract once; do not
add V1/V2 dual handling.

## File changes

Action authoring/runtime:

- `packages/core/src/edits/types.ts`;
- `packages/core/src/edits/recorder.ts`;
- `packages/core/src/edits/normalize.ts`;
- `packages/core/src/edits/properties.ts`;
- `packages/core/src/actions/read-facade.ts`;
- `packages/core/src/actions/commit-edit-batch.ts`;
- `packages/core/src/actions/types/context.ts`;
- `packages/core/src/actions/types/handlers.ts`;
- `packages/core/src/storage/action-runs/types.ts` and providers;
- `packages/action-worker/src/action-execution/context.ts`;
- `packages/action-worker/src/action-execution/edits-commit.ts`;
- `packages/action-worker/src/action-execution/results.ts`.

Runtime object API wiring:

- `packages/core/src/runtime/sixb.ts`;
- `packages/core/src/runtime/types.ts`;
- `packages/core/src/objects/context/types.ts`;
- `packages/core/src/objects/sdk/object-set.ts`;
- `packages/core/src/objects/sdk/object-handle.ts`;
- `packages/core/src/objects/service/object-service.ts`;
- `packages/core/src/objects/service/link-service.ts`;
- `packages/core/src/objects/object/upsert.ts`;
- `packages/core/src/objects/object/upsert-batch.ts`;
- `packages/core/src/objects/link/upsert.ts`;
- `packages/core/src/objects/link/upsert-batch.ts`;
- `packages/core/src/objects/link/remove.ts`.

Create `packages/core/src/objects/materializer-adapter.ts` for runtime translation/result mapping.

Delete:

- `packages/action-worker/src/action-execution/commit-events.ts`;
- `storage/sqlite/src/action-run-commit-diff.ts`;
- `storage/pg/src/action-run-commit-diff.ts`;
- Action-specific physical plan/application code extracted from `packages/core/src/edits/validation.ts`;
- every `emitLocalCommitEvents` call.

## Final EditBatch authoring contract

Replace the current contract in place:

```ts
interface EditBatch {
  readonly version: 1
[...]

PR #257, opened 2026-07-26
core: route Actions and the runtime object API through durable edits

demattosanthony:
Phase 3 of the ontology materializer stack (#234). Actions and the runtime object/link API become
thin producers of `OntologyEditCommit`, the EditBatch contract is replaced in place, and telemetry
writes move onto the materializer. Net **-5,277 lines**.

## The need

Before this change there were two mutation architectures. Actions ran their own planner and
persisted a duplicate semantic diff; runtime object and link writes were event-first, appending a
domain event and letting providers apply it as a storage command. Each path had its own validation,
its own diff, and its own event construction, so a semantic change had to be made twice and could
silently diverge.

| Concern | Old | Consequence |
| --- | --- | --- |
| Authority | Action planner and runtime writer both computed it | Two sources of truth for the same transition |
| Events | `events.append()` published inline, before state was readable | A subscriber could observe an event whose state was not committed |
| Durability | Broker failure surfaced to the caller mid-write | A broker outage failed a write that had already applied |
| Action runs | Stored `commit_id` plus a physical diff | A second commit ledger to keep consistent |

```mermaid
sequenceDiagram
    participant C as Caller
    participant R as Runtime / Action planner
    participant P as Provider
    participant B as Broker

    C->>R: upsert / action edits
    R->>R: validate + diff (per-path logic)
    R->>B: events.append(...)
    B-->>R: published
    R->>P: apply event as storage command
    P-->>R: rows written
    R-->>C: ok
    Note over R,B: Event published before state is readable
    Note over R,B: Broker failure fails the write
```

## The fix

1. Every Action commit and every typed or dynamic runtime mutation goes through
   `materializer.edits.commit`. No mutation path appends events or writes providers directly.
2. One final `EditBatch` contract replaces the old one — no V1/V2 branch. The recorder exposes
   `create` and `byId(...).update/unset/reset/delete/restore/link/unlink/resetLink`.
3. Domain events become durable outbox rows written inside the same transaction as the state, then
   published after commit. A broker outage delays delivery instead of failing the write.
4. Action commits capture the exact reads a run depended on and fail if that state changed.
5. `ActionRunStorage` no longer stores a commit pointer or a duplicate diff; the ontology commit
   ledger is authoritative and resume resolves it by exact `{ actionRunId }` origin.

```mermaid
sequenceDiagram
    participant C as Caller
    participant A as Runtime / Action adapter
    participant M as Materializer
    participant P as Provider
    participant D as Outbox dispatcher
    participant B as Broker

    C->>A: upsert / action edits
    A->>M: edits.commit(operations, expected reads)
    rect rgb(238, 246, 238)
        Note over M,P: one serializable transaction
        M->>M: resolve, validate, diff once
        M->>P: effective state + authoritative commit + outbox rows
    end
    P-->>M: committed
    M-->>A: persisted result
    A-->>C: ok (state readable)
    A->>D: drain()
    D->>B: publish envelopes
    Note over D,B: Failure reschedules the rows and the commit stands
```

### Why this is better

| | Before | After |
| --- | --- | --- |
| Mutation paths | Action planner + event-first runtime | One materializer |
| Event timing | Published before state was readable | Durable outbox row, published after commit |
| Broker outage | Failed the caller's write | Delays delivery only |
| Event order in a commit | sha256 envelope id, effectively random | `commit_ordinal`, so a link never precedes its object |
| Batch item failure | Could commit half of a multi-operation item | Rolls the item back, reports every position failed |
| Validation error class | Single threw `MaterializationValidationError`, batch threw `OntologyValidationError` | Both catchable as `OntologyValidationError` |
| Action run record | Commit pointer + duplicate diff | Ledger is authoritative |
| Mutation-path size | baseline | **-5,277 lines** |

## Review fixes included

The cutover was reviewed before this PR. Six defects were found and fixed here, each with a test
that fails without its change:

| Fix | What was wrong |
| --- | --- |
| Group-atomic batch items | A cardinality-one reassignment whose `link.upsert` failed still committed its `link.delete`, destroying the edge while reporting an item error |
| Writeback read fencing | The writeback read facade had no recorder, so reads that informed an external call were never fenced |
| Provenance-less reads | Object reads of such rows recorded nothing while link scopes recorded a fingerprint that could never match |
| Empty-update guard | `update({})` committed as `unchanged`, so a conditionally built update reported success having written nothing |
| Outbox settlement | `drain()` and the poll loop shared one in-flight slot, so one batch went unsettled and was re-delivered after lease expiry |
| Publication order | Claim order tiebroke on the envelope hash, so a link could be delivered before the object it references |

Full review and the rationale for what was deliberately left alone:
`.local/drafts/materializer-phase-3-review.md`.

## Rollout notes

- **Schema.** `001-initial-schema` is rewritten in place for both PostgreSQL and SQLite, per the
  clean-break plan in #237. Any database that already ran the old `001` will fail its checksum
  check on boot. This is intended at this stage of the stack — there is no upgrade path and none is
  planned before the release notes in #236.
- **Breaking: Action edits facade.** `objects(Type).upsert(...)` is gone. Use `create(...)` or
  `byId(id).update(...)`. An action synchronizing a record that may or may not exist must read
  first and branch. The runtime `sixb.objects(...).upsert` is unchanged.
- **Breaking: link properties.** A link upsert writes the complete property set. Properties left
  out are cleared, and required link properties must be present on every write.
- **Breaking: batch duplicates.** A batch may not claim one identity twice with different content.
  An exact repeat collapses onto the first item; a differing repeat is rejected at its own position.
- **Behavioral: error class.** `MaterializationValidationError` now extends
  `OntologyValidationError`, so existing `instanceof OntologyValidationError` handling keeps
  working across both single and batch writes.
- **Not in this phase.** Nothing starts the outbox dispatcher — #236 assigns hosting to the server,
  standalone workers, worker groups, and a dedicated outbox worker, and specifies that the `Sixb`
  constructor starts no hidden polling. Until then a broker outage leaves facts pending. Two latent
  dispatcher bugs were fixed here so that hosting can be wired safely.
- **Transitional.** Foreign-key projections that emit one dataset row per edge are degraded until
  #235 moves projection workers onto `projections.replace`.

## Tests

| Suite | Result |
| --- | --- |
| `bun run test` | 3,022 pass, 2 skip |
| `bun test storage/sqlite/tests/` | 241 pass |
| PostgreSQL E2E (`storage/pg`) | 209 pass |
| `bun run typecheck` / `build` / `check` | pass |
| `bun run test:publish` | 44 publishable packages verified |
| `bun run generate:client` | no drift |

Added or restored:

- Group rollback: a failed cardinality-one assignment leaves the original edge intact, alone and
  alongside a succeeding item in the same batch.
- Writeback fencing: a non-subject object read in writeback and mutated before the commit fails the
  run.
- Outbox: a poll-loop publication overlapping a concurrent drain settles; 20 concurrent drains cost
  at most 4 claim transactions.
- Ordering: the three tests that sorted published events now assert the order.
- Error parity: single and batch writes report the same catchable class and message.
- Link properties: replacement clears omitted properties; a required property is rejected when
  omitted, leaving the stored edge unchanged.

The concurrent cardinality-one test previously asserted only that at least one item succeeded. It
now asserts that exactly one edge survives and that it belongs to an assignment that reported `ok`,
so a losing item cannot leave state behind.

Not run locally: the remaining E2E matrix (ducklake, blob-s3, bullmq, nats, redis, cli). Nothing in
this branch touches them.


quentinnippert:
@demattosanthony 
## Updates

### Correctness

- Rules now use events as wake-ups and always read the current committed state.
  - Example: a delayed `Alice → Alicia` event evaluates `Alicia`, not the stale payload.
- Action commits now:
  1. check exact replay;
  2. lock the Action run (`FOR UPDATE` on PostgreSQL);
  3. replay inside the transaction;
  4. materialize only if still absent.
- A crashed Action can resume even when its commit deleted its subject.
- `operationGroups` are compiled into validated, contiguous execution units.
  - Example: `delete old link → create new link` fully rolls back if creation fails.

### Performance

```text
read-set → paginated batch load → ordered in-memory apply
```

- Object/link edits no longer perform per-operation reads.
- Telemetry loads object state and existing points in batches.
- Batch tests cover `1`, `1000`, and `1001` items.

### Events and outbox

- Public `events.append()` can no longer author object, link, or telemetry facts.
- Authorable, stored, and Materializer event contracts are separated and typed.
- Server/OpenAPI schemas validate exact ontology event payloads.
- `EventOrigin` variants are named without changing their persisted format.
- Outbox publication is non-blocking, coalesced, and tracked during shutdown.
- `commitOrdinal` is correlation within a commit, not a delivery-order guarantee.

### Boundaries

- Public Materializer/outbox exposure was replaced with the internal `OntologyMutationRuntime` capability.
- Batched telemetry uses the same public missing-object error contract as other object APIs.

### Deferred

- Persisting `writeback.ontologyReadSet`.
- Action change history in API/UI (`#236`).

quentinnippert (approved):

PR #277, opened 2026-07-30
app: bound the Tailwind CSS build

quentinnippert:
`runCompile()` spawned the Tailwind CLI, then `await proc.exited` — no bound, no kill, and stderr
read only after the wait.

```ts
const proc = Bun.spawn(args, { cwd, stdout: "ignore", stderr: "pipe" })
const exitCode = await proc.exited          // waits forever if the child never exits
```

## Three defects in two lines

| | Effect |
|---|---|
| No bound | `compile()` hangs forever. `sixb dev` never becomes ready; a CI job dies at its own wall clock with no failing test named |
| No `kill()` | the child outlives the function — the runner reaps it as an orphan, which is how these hangs are recognizable after the fact |
| stderr read after `exited` | a child that fills the ~64 KB pipe blocks on write while the parent waits for the exit that write is blocking. **Deadlock, not a slow build** |

The third one is latent rather than active: measured **36 bytes** of stderr on the Atlas stylesheet
today. The ordering is wrong regardless, and it costs nothing to read concurrently.

## The change

`timeoutMs`, default **two minutes** — generous for a cold build on a loaded runner, and an option
because a large app may legitimately need longer. The error names the input, the bound, and the knob:

```
[SixbAtlas] Tailwind CSS build did not finish within 120000ms and was stopped.
Check that '…/index.css' compiles, or raise 'timeoutMs'.
```

## What this does not claim

It is **not** proven to fix the `Unit tests` job that has been dying at its 15-minute timeout inside
`packages/atlas/tests/atlas-app.test.ts`. Evidence against that reading, produced while teeth-testing
this change: with the kill removed, the new test fails at **5.05 s** via Bun's per-test timeout rather
than hanging. So Bun *can* interrupt a stuck `await proc.exited` — which means the job that runs 15
minutes without any per-test timeout firing is stuck somewhere else. The remaining suspect on that path
is `ensureBuiltInUiDevBundle()`'s `await import(htmlPath)`, Bun's in-process HTML bundler.

This lands on its own merits: an unbounded wait on a child process, an unreaped child, and a pipe read
in the wrong order are defects whether or not they are today's outage.

## Verification

- `bun --filter @sixb/app typecheck` · `bun run build:types` · `bun run check`
- `bun test packages/app/tests/` — 37 pass / 1 fail, and that one failure is **pre-existing on `main`**
  (36 pass / 1 fail there; this change adds the passing test)
- Teeth: with `proc.kill()` removed the new test fails instead of passing

### DIFF
--- a/packages/app/src/tailwind.ts
+++ b/packages/app/src/tailwind.ts
@@ -25,6 +25,17 @@ export interface TailwindCssCompilerOptions {
   minify?: boolean
   /** Debounce for `schedule()` rebuilds, in milliseconds. */
   debounceMs?: number
+  /**
+   * How long one Tailwind CLI run may take before it is killed and the build
+   * reported as failed. Defaults to two minutes, generous enough for a cold
+   * build on a loaded CI runner.
+   *
+   * A bound is not optional here: the CLI runs as a child process, and a child
+   * that never exits makes `compile()` hang forever — outside any test runner's
+   * per-test timeout, so the whole job dies at its wall clock with no failing
+   * test named, leaving orphaned `bun` processes behind.
+   */
+  timeoutMs?: number
   /** Called when a `schedule()`d rebuild fails. Defaults to `console.error`. */
   onError?: (error: Error) => void
 }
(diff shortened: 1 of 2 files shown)


quentinnippert:
Closing: this was opened on a diagnosis I then invalidated myself. Teeth-testing the change showed Bun's per-test timeout already interrupts a stuck `await proc.exited` at 5s, so the job that runs 15 minutes with no per-test timeout firing is stuck elsewhere. And the pipe-deadlock leg is latent, not active — measured 36 bytes of stderr on the Atlas stylesheet.

The bounded-wait-with-a-legible-error item is real and moves into the `runtime: boot, health, and honest degradation` lot, which already owns that family (`renderCliError`, `close?()` on the provider contracts, `requireCapability`).

PR #319, opened 2026-08-06
release: publish packages independently

demattosanthony:
## Summary

- replace the repository-wide lockstep version with explicit, per-package release intent
- add a registry-backed `release:plan` command that shows exactly which local versions are missing
- publish only those versions, while preserving the existing dependency order, artifact gate, resumability, and preview-tag policy
- prepare the packages changed since `0.1.0` for their next release

## Why

The existing release script required every publishable workspace to share one version. With 47 packages at different stages, releasing one connector or worker therefore meant bumping and republishing unrelated packages.

This keeps the current lightweight Bun-based release flow, but makes each package manifest authoritative for that package. Before the first registry write, the publisher reads registry state for the entire workspace, removes versions that already exist, and verifies that every internal dependency version is already available or appears earlier in the publish plan.

That gives us selective releases without introducing a larger release-management system.

## Release behavior

The release flow is now:

```bash
bun run release:plan -- --tag next
bun run release
bun run release:publish -- --tag next
```

- `release:plan` is read-only and prints the exact package/version set.
- `release` still performs the clean build and validates every publishable artifact.
- `release:publish` publishes only missing versions in topological order.
- Re-running an interrupted release skips immutable versions already on the registry.
- Promotion commands include only packages from the selective release, including packages staged during an earlier partial run.
- `0.0.x` releases remain restricted to the `next` tag.

The publish gate no longer enforces a shared version, but it still requires every package to declare a version and retains all existing package-boundary, export, tarball, and dependency-cycle checks.

## Prepared package versions

| Package | Version | Reason |
| --- | ---: | --- |
| `@sixb/core` | `0.1.1` | reusable agent tool definitions |
| `@sixb/agent-worker` | `0.1.1` | execute tools selected by an agent |
| `@sixb/connector-github` | `0.1.1` | expanded people and membership APIs |
| `@sixb/sync-worker` | `0.1.1` | handle empty initial snapshots |
| `@sixb/pg` | `0.1.1` | handle empty initial snapshots |
| `@sixb/sqlite` | `0.1.1` | handle empty initial snapshots |
| `@sixb/connector-exa` | `0.1.0` | first publication |
| `@sixb/connector-google` | `0.1.1` | align the manifest and changelog with the already-published Gmail release |

The current registry-backed plan selects seven packages. Google and the other 39 current versions are already on the registry and are skipped.

## Validation

- `bun test scripts/tests/` — 55 passed
- `bun run typecheck`
- `bun run check`
- `bun run release` — clean build and publish-artifact validation for all 47 packages
- `bun run release:plan -- --tag next` — seven packages selected; nothing published
- `git diff --check`

No packages were published while preparing or validating this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,24 @@
 # Changelog
 
-All Sixb packages ship on one version. A release publishes all of them or none.
+Sixb packages are versioned independently. Each release entry names the packages that shipped.
+
+## 2026-08-05
+
+### 0.1.1
+
+- `@sixb/core` and `@sixb/agent-worker`: define reusable agent tools and run the tools selected by
+  each agent.
+- `@sixb/connector-github`: add users, memberships, members, invitations, and outside collaborators.
+- `@sixb/sync-worker`, `@sixb/pg`, and `@sixb/sqlite`: handle empty initial snapshots without
+  creating an unusable dataset version.
+
+### `@sixb/connector-exa` 0.1.0
+
+Add bounded web search and fetch tools as a new connector package.
+
+### `@sixb/connector-google` 0.1.1
+
+Add the complete typed Gmail v1 surface.
 
 ## 0.1.0
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -185,18 +185,21 @@ Prefer targeted checks while iterating, then broader checks before review.
 
 ## Versioning
 
-Every publishable package shares one version and ships together — a release publishes all of them or
-none. `bun run test:publish` fails if they drift apart, so a version bump touches every manifest.
+Publishable packages are versioned independently. A package manifest is its release intent: bump the
+packages that should ship and leave unrelated manifests alone. The publisher reads registry state
+before its first write, publishes only local versions that do not exist yet, and keeps dependency
+order across that smaller plan.
 
 The `0.0.x` line is for public validation. Publish those versions only under npm's `next` tag and
 increment the patch for every new artifact: npm versions are immutable. The `latest` tag is reserved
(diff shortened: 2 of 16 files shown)


quentinnippert:
@demattosanthony thoughts on this?

One thing worth aligning on before release: Bun publishes `workspace:*` as an exact version, so changing `core` also ties its dependents to the release.

options:
- Keep exact versions and release packages together
- Use `workspace:^` for independently compatible packages
- Use a hybrid model: exact for `core/internal`, ranges for public APIs, and peers for plugins/providers

I like the hybrid model, it probably fits our independent versioning strategy best. It needs a one-time migration, then keeps future releases selective.

demattosanthony:
@quentinnippert agreed — the hybrid model is the best fit for the independent versioning work here. I implemented it in #335, stacked on this PR.

The resulting contract is:

- public package APIs use `workspace:^`;
- connectors/providers expose core as a compatible peer and keep `workspace:*` only as their development dependency;
- workers, server, orchestration, and storage packages that import `@sixb/core/internal/*` peer on core exactly;
- the CLI pins the core/internal runtime it owns exactly, while its public UI/client dependencies stay compatible.

I also made this enforceable rather than leaving it as a manifest convention. The publish gate checks every workspace edge and audits the exact cohort against real internal imports. The release planner compares already-published dependency fields/ranges with the current workspace before skipping a version, so it requires bumps for changed exact contracts while allowing caret consumers to stay selective.

This is the one-time migration cost: the live plan currently has 45 packages to publish and two already current. After that, compatible packages can remain untouched across core patch releases, while the internal cohort moves together when required.

quentinnippert (approved):

PR #355, opened 2026-08-10
storage: add durable AI usage attribution

demattosanthony:
## Summary

> Stack: #354 → **#355** → #356 → #357 → #358

This is PR 2 of 5 in the Phase 1 AI usage accounting stack. It takes the provider-neutral contract
from the previous PR and makes it durable in the composite in-memory, SQLite, and PostgreSQL storage
providers. It also snapshots requester group attribution when parent runs are admitted.

The existing run-level usage projections and API behavior remain unchanged in this PR. That keeps
this additive storage rollout independently deployable; the final stack PR removes those projections
after both capture paths and ledger-backed reads exist.

## Why durable attribution is required for token budgets

A future budget may apply simultaneously to:

- the project;
- the requesting principal;
- every group the requester belonged to when the work was admitted.

Resolving groups when a model call finishes would move historical usage when membership changes.
Using only a token-restricted authorization context would also let a caller avoid a shared group
budget by presenting a narrower token.

This PR therefore snapshots all durable group memberships at admission and carries that immutable
snapshot through the parent run. Later capture PRs copy it onto every model-call ledger record.

### Durable AI usage storage

- Wires `Storage.aiUsage` into the composite storage surface.
- Adds transactional in-memory snapshots and rollback.
- Adds SQLite and PostgreSQL append-only ledger implementations.
- Writes model-call rows and requester-group rows atomically.
- Enforces idempotent replay and duplicate-record behavior.
- Adds execution/project/time/principal/group/provider-response indexes.
- Implements ordered batch summaries efficiently:
  - one in-memory ledger pass;
  - one set-based SQL read;
  - unavailable entries preserved in input order.
- Runs the same provider contract against all three implementations.

### Admission-time attribution

- Adds immutable `requesterGroupIds` to conversation parent runs.
- Adds immutable `requesterGroupIds` to workflow parent runs.
- Resolves memberships from durable auth storage rather than authorization-token group subsets.
- Makes scoped runtime identity authoritative over caller-supplied principal fields.
- Snapshots user and service-account memberships.
- Gives system principals an empty group snapshot.
- Treats an agent retry as a fresh admission with a fresh membership snapshot.
- Uses the workflow parent run as the only attribution source for workflow agent nodes.

### Migration 004

Adds one combined additive migration for each SQL provider:

`004-ai-usage-accounting-foundation`

It:

- creates the ledger and group tables;
- creates idempotency and query indexes;
- adds `requester_group_ids` to agent parent runs;
- adds `requester_group_ids` to workflow parent runs.

The original implementation had separate ledger and attribution migrations. They are combined here
because trustworthy usage storage and stable attribution are one operational foundation, not two
independent rollout states.

## Rollout behavior

This PR is additive:

- workers do not write the ledger yet;
- existing run-row usage aggregates continue to be written;
- HTTP and Atlas continue to read existing run summaries;
- no historical usage is removed.

That means migration 004 can land before worker capture without causing a user-visible gap.

## What is intentionally not in this PR

- No `streamText` or `generateText` lifecycle capture.
- No accounting failure gating.
- No ledger-backed API reads.
- No usage-column drop.
- No pricing, budget counters, configuration, or enforcement.

## Phase 1 stack

1. Accounting contract and standalone in-memory model.
2. **This PR:** durable storage and immutable requester attribution.
3. Conversation model-call capture.
4. Workflow agent-node model-call capture.
5. Ledger-backed API cutover and removal of legacy run projections.

## How this supports later budgets

This PR establishes the durable dimensions future counters need:

- project ID;
- principal type and ID;
- stable group IDs;
- provider/model/response identity;
- occurrence time;
- normalized and raw usage.

A later monthly-budget transaction can rate one immutable usage fact and increment project,
principal, and group counters without re-resolving identity or membership.

After Phase 1, separate work will add effective-dated pricing, reconciliation, monthly counters,
budget definitions, and pre-call/between-step enforcement. None of those concerns are coupled into
this storage hot path yet.

## Review guide

Suggested order:

1. `packages/core/src/auth/attribution.ts`
2. agent/workflow admission changes
3. combined migration 004 for one provider, then parity-check the other
4. SQLite/PostgreSQL `*ai-usage-storage.ts`
5. composite transaction wiring
6. provider and migration tests

Questions to keep in mind:

- Can token-scoped groups evade attribution?
- Does a membership change mutate an existing run snapshot?
- Are workflow child nodes prevented from inventing separate attribution?
- Are model-call/group rows atomic?
- Does replay return the original record without changing attribution?
- Are project and execution identities isolated in every summary query?
- Does the batch path avoid N reads?

## Validation

```bash
bun run typecheck
bun run check
bun test packages/core/tests/ai-usage-storage.test.ts \
  packages/core/tests/agent-dispatch.test.ts \
  packages/core/tests/scoped-sixb.test.ts \
  packages/core/tests/workflow-request.test.ts \
  packages/core/tests/agents-storage.test.ts \
  packages/core/tests/workflow-runs.test.ts
bun test storage/sqlite/tests/sqlite-ai-usage-storage.test.ts \
  storage/sqlite/tests/sqlite-agent-storage.test.ts \
  storage/sqlite/tests/sqlite-workflow-run-storage.test.ts \
  storage/sqlite/tests/sqlite-storage-migrations.test.ts
bun --filter @sixb/pg test:e2e
```

Validated results include 105 focused core tests, 58 focused SQLite tests, and all 183 PostgreSQL E2E
tests. Core, SQLite, PostgreSQL, and server builds pass.

### DIFF
--- a/packages/agent-worker/tests/worker.test.ts
+++ b/packages/agent-worker/tests/worker.test.ts
@@ -1223,6 +1223,7 @@ describe("AgentWorker", () => {
       executionId,
       workflowId: workflow.id,
       input: { query: "alpha" },
+      requesterGroupIds: [],
     })
     await runs.start({
       id: runId,
@@ -4144,6 +4145,7 @@ describe("AgentWorker", () => {
       threadId,
       agentId: "removed-agent",
       triggerMessageId,
+      requesterGroupIds: [],
     })
     await sixb.queues.agents.enqueue({
       projectId: PROJECT_ID,
--- a/packages/core/src/agents/request.ts
+++ b/packages/core/src/agents/request.ts
@@ -1,6 +1,7 @@
 import { randomUUID } from "node:crypto"
 import type { Principal } from "../auth"
 import { SYSTEM_PRINCIPAL } from "../auth"
+import { snapshotRequesterGroupIds } from "../auth/attribution"
 import { assertAuthorized } from "../authorization"
 import type { FileRef } from "../blob-storage"
 import { createAgentExecutionRecord } from "../execution/agent"
@@ -38,7 +39,7 @@ export interface RequestAgentRunInput {
   readonly title?: string
   /** Explicit id for the trigger (user) message. Defaults to a generated id. */
   readonly messageId?: string
-  /** Owner principal for a new thread. Defaults to system for an auth-disabled execution. */
+  /** Caller principal for privileged integrations; scoped runtimes use their authorization principal. */
   readonly principal?: Principal
 }
 
@@ -74,8 +75,9 @@ export async function requestAgentRun(
 
   const agents = requireAgentStorage(runtime)
   const projectId = runtime.projectId
-  const principal = input.principal ?? SYSTEM_PRINCIPAL
-
+  // A scoped runtime is authoritative for caller identity. `input.principal` remains available to
+  // privileged server integrations that have already authenticated their request.
+  const principal = runtime.authorization?.principal ?? input.principal ?? SYSTEM_PRINCIPAL
   const { thread, createdThread } = await resolveThread(agents, {
     projectId,
     agentId: agent.id,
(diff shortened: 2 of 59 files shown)


quentinnippert:
@demattosanthony maybe we merge #382 first. There's a conflict on this one with the execution layer.
Once merged I'll rework this one

quentinnippert (approved):

Issue #274, opened 2026-07-29
Idea: one error model


quentinnippert:
One failure can surface in three places, under three shapes, and none of them carries an identity a
consumer can branch on. The class name is the de-facto contract, which is why there are a hundred of
them.

## Where we are

| | Today |
| --- | --- |
| Error classes | **104** (63 in `@sixb/core`), a few carrying a `code`, most not |
| Failure shapes in storage | **5** — `ActionRunFailure` / `SyncRunFailure` / `PipelineRunFailure` (near-identical `{ name?, message }`), a bare `error?: string` for workflow, agent and webhook runs, and `errorMessage?: string` for projection runs |
| Escalation channels | **4** — `throw`, the run record's error column, `console.error`/`warn` (50 sites), and `onError` |
| Stable machine-readable code | none, anywhere |

An operator asking "what broke, is it worth retrying, and did anyone get told" has to answer it
differently per primitive.

### 1. One error type, not a hundred

```ts
class SixbError extends Error {
  readonly code: SixbErrorCode      // the contract
  readonly retryable: boolean
  readonly details?: JsonValue
  readonly cause?: unknown
}
```

Codes are namespaced and enumerated: `workflow.run.node_failed`, `storage.serialization_conflict`,
`auth.session_expired`. Subclasses survive **only** where a consumer needs `instanceof` for control
flow — validation, authorization, conflict, timeout, provider. The `code` travels; the class is a
convenience.

### 2. One failure record

```ts
interface SixbFailure {
  readonly code: SixbErrorCode
  readonly message: string
  readonly retryable: boolean
  readonly at: string
  readonly details?: JsonValue
  readonly causeChain?: readonly { readonly name: string; readonly message: string }[]
}
```

Stored verbatim in every run table. Atlas renders one component, a consumer writes one handler, and
the three near-identical twins plus the four string columns go away.

### 3. Errors are values at the boundary, exceptions inside

Stated once, so it stops being decided per module by whoever wrote it:

> "forbidden", "not found", "already running" are **values**. "the database is down" is an **exception**.

Public entry points that can fail for expected reasons return a discriminated result —
`RequestActionResult` already has that shape. Framework internals throw.

### 4. One escalation channel

Today an operator gets `console.error` (invisible in production), `onError` (two context variants), or
a run column (per primitive). Instead, everything Sixb wants to tell an operator goes through one typed
sink:

```ts
observe(event: SixbRuntimeEvent)
[...]

PR #406, opened 2026-08-19
broker: bound a Redis command so one stuck reply cannot stop all publication

### LINKED ISSUE #217: broker: Redis append can block all event publication
A `RedisBroker` append remained pending after Redis had successfully stored a batch of 294 events (~9.4 MB). The shared command queue then stopped all further event publication.

In ADN, this left a projection running, an action stuck in `commit` after its PostgreSQL commit had completed, and later actions queued indefinitely. Redis and PostgreSQL remained healthy.

Environment: Bun 1.3.8, Redis 7.2.

yashksaini-coder:
## Summary

Every broker append and read shares one Redis client and is serialized through
`commandQueue`, and the lock was only released in the `finally` of the awaited
operation. Nothing bounded that operation, so a command that was sent and never
answered held the lock forever: all further event publication stopped, and
`close()` waits on the same queue, so the process could not shut down and
self-heal either. That is the outage in #217.

Each operation on the shared client is now bounded by
`connection.commandTimeoutMs` (30 seconds by default). Bun's `connectionTimeout`
covers connecting and `idleTimeout` covers an idle socket, but neither bounds a
command that is already in flight, so the bound has to be ours.

A timed-out client is discarded rather than reused. Its missing reply may still
arrive, and a late reply on a shared connection can be matched to the wrong
command, so keeping it would trade a visible hang for silent corruption.
`connect()` already opens a fresh client when there is none.

The subscription path already had its own watchdog around `XREAD BLOCK`. This
closes the same gap on the shared command path.

## Linked issue

Fixes #217.

## Scope

Two files of behaviour: `broker/redis/src/connection.ts` for the bound, and one
line in `redis-broker.ts` validating the new option through the existing
`positiveInteger`, alongside the other numeric options.

Deliberately left out:

- **Chunking large appends.** The reported batch was ~9.4 MB in one `EVAL`, and
  the outbox dispatcher batches up to 1000, so roughly 3x larger is reachable
  today. But `appendBatch` uses one Lua call on purpose to keep the XADDs and
  dedupe-key writes atomic across the batch, and chunking trades that away. That
  felt like your call.
- **Removing the shared command lock.** Serializing every read and append in the
  process is a real throughput ceiling, and the comment justifies it by response
  interleaving that a correct RESP client should already prevent. Removing it
  would not have fixed this bug, though, since an unbounded command still hangs
  its own caller. Happy to open a separate issue.
- **Attributing the timeout error to a stream.** The timeout rejects outside
  `appendBatch`'s `catch`, so unlike every other error in the package it names no
  stream. Fixing it means threading a label through `useCommandClient` at 12 call
  sites, which seemed too much for a bug fix. Worth a follow-up.
- **A distinct timeout error class**, because #274 is consolidating error classes
  rather than adding them. In that model this error wants `retryable: true`, and
  it is worth remembering that a timeout is *indeterminate* rather than failed.

## Validation

`bun run check`, `bun run typecheck`, `bun run build`, `bun run test:publish`,
`bun run generate:client` + `git diff --exit-code`, and `bun run test:ci`
(3804 pass / 0 fail). `bun --filter @sixb/broker-redis test:e2e` passes 42/42
against Redis 7.2 from the package's `docker-compose.yml`.

Four unit tests use the existing injectable client factory, so they need no
Redis and add ~150ms. Each was proven by reverting the fix: without it all four
hang until the per-test timeout. That check also caught a fifth test of mine that
passed either way -- closing too early aborts the command through `assertOpen()`
instead -- so it now waits until the command is genuinely in flight. The repro is
noted in the file. `useCommandClient` had no coverage before this.

## Notes

- `commandTimeoutMs` must be a positive integer. The unvalidated version was a
  trap: `Number(process.env.X)` on an unset variable is `NaN`, `setTimeout`
  coerces that to 1ms, and every command fails; a negative value silently
  restored the #217 hang. There is deliberately no "wait indefinitely" escape --
  a large number covers the slow-Redis case without offering a way to switch the
  guard off by accident.
- A timeout is **indeterminate**: Redis may apply the command afterwards. The
  outbox path is safe because every envelope carries `idempotencyKey` and
  `dedupeTtlMs` (120s) is longer than the bound, so a retry is deduped. That
  `commandTimeoutMs < dedupeTtlMs` invariant is currently undocumented -- worth
  saying out loud somewhere if you agree.
- An abandoned operation that later *succeeds* is logged, not just one that
  fails. That is the case an operator needs to explain a duplicate.
- One behaviour change to be aware of: a Redis failover longer than the bound no
  longer completes transparently through Bun's reconnect and offline queue.
  README updated for all of the above.
- `connect()` is intentionally outside the bound; Bun's `connectionTimeout` does
  cover connecting.


yashksaini-coder:
Ready for review.

One thing worth flagging: #364 changed the same files shortly before this branch point, but it fixed the subscription path. This is the shared command path, and #364 is already in the base here.

CI needs maintainer approval to run on a fork PR. Locally green: lint, typecheck, build, `test:publish`, `generate:client` (no contract change), `test:ci` 3805/0, and broker e2e 42/42 against Redis 7.2.

yashksaini-coder:
@quentinnippert Will u be taking this over, shall I take a different issue, or would like to collaborate and co-author on this

quentinnippert:
@yashksaini-coder Just added some stuff, I appreciate your contribution a lot. We can be co-author if that's fine with you

yashksaini-coder:
> @yashksaini-coder Just added some stuff, I appreciate your contribution a lot. We can be co-author if that's fine with you

Great we can be co-authors, I'm glad you took interest in my apporach and pr, I will be doing more approved issues, as well raise few of my own.

PR #407, opened 2026-08-19
core: release an action run wait subscription that arrives after the wait settled

### LINKED ISSUE #405: core: a timed-out action run wait leaks its event subscription
`waitForActionRun` releases its event subscription from `cleanup()`, but assigns the unsubscribe handle inside the `.then()` of an asynchronous `events.subscribe()`.

When the wait settles first — on timeout, or on abort — `cleanup()` runs while the handle is still undefined, and the subscription that arrives afterwards is never released. The wait itself returns correctly, so nothing surfaces.

Under `@sixb/broker-redis` each subscription holds a dedicated Redis client running a blocking XREAD pump, so every timed-out `requestAction(...).wait()` leaks a connection for the life of the process.

`packages/core/src/actions/request.ts`, in the `subscribe().then(...)` chain at the end of `waitForActionRun`. The equivalent path in `@sixb/client` subscribes synchronously and is not affected.

yashksaini-coder:
## Summary

`waitForActionRun` releases its event subscription from `cleanup()`, but assigns
the unsubscribe handle inside the `.then()` of an asynchronous
`events.subscribe()`. When the wait settles first -- on timeout or on abort --
`cleanup()` runs while the handle is still undefined, and the subscription that
arrives afterwards is never released.

Under `@sixb/broker-redis` each subscription holds a dedicated Redis client
running a blocking XREAD pump, so every timed-out `requestAction(...).wait()`
leaks a connection for the life of the process.

Release the subscription when it resolves after the wait has settled. `settled`
already exists in that closure, so no new state is needed.

## Linked issue

Fixes #405.

## Scope

One behavioural change in `packages/core/src/actions/request.ts`, plus first
coverage for `waitForActionRun`.

The `.catch` on the same chain is also fixed. Once the wait has settled both
halves of `rejectWith` are inert -- `cleanup()` returns early and `reject()` no
longer changes the promise -- so a `subscribe()` that rejected late, or a release
that threw, vanished with no trace at all. It is now reported. That is one extra
guard on the same path rather than a separate concern, and it matters more now
that a bounded broker command gives `subscribe()` a new way to fail.

## Validation

`bun run check`, `bun run typecheck`, `bun run build`, `bun run test:publish`,
`bun run generate:client` + `git diff --exit-code`, and `bun run test:ci`
(3804 pass / 0 fail).

The new test holds `subscribe()` open past the timeout with a deferred, so the
race is deterministic rather than timing-dependent. Proven by reverting the fix:
the wait still times out, but nothing unsubscribes and the test fails on the
assertion. The repro is noted in the file.

## Notes

- The equivalent path in `@sixb/client` subscribes synchronously and does not
  have this bug, so this is the only affected site.
- `waitForActionRun` had no test anywhere in `packages/core/tests/` before this.
- No version bump included -- `@sixb/core` bumps cascade to exact-peer consumers,
  so that felt like your call rather than something to fold into a bug fix.


yashksaini-coder:
Ready for review. Fixes #405.

CI needs maintainer approval to run on a fork PR. Locally green: lint, typecheck, build, `test:publish`, `generate:client` (no contract change), and `test:ci` 3800/0.

Independent of #406, so the two can land in either order.

quentinnippert:
@yashksaini-coder thank you for your contribution! 🙏

yashksaini-coder:
> @yashksaini-coder thank you for your contribution! 🙏

Can I take some more approved issues

Issue #269, opened 2026-07-28
Idea: capability-scoped shared links for external access


demattosanthony:
## Outcome

Let an application grant temporary access to one exact object through a transferable bearer link,
without requiring the recipient to have a Sixb account.

The recipient opens an ordinary application page. That page keeps using the standard Sixb client,
queries, and Action mutations; Sixb binds those requests to the link's narrow delegated authority.

```text
exact ObjectRef
  -> issued grant
  -> bearer link
  -> isolated shared session
  -> ordinary page and canonical APIs
  -> exact objects, link paths, and Actions selected by the Share
```

The framework owns credentials, sessions, authorization, OIDC isolation, and attribution. The
application owns the page, link delivery, and domain behavior.

## Developer experience

Share definitions are auto-discovered from `shares/`. They reuse the existing grant vocabulary;
there is no separate public `graph` or shared-resource model.

```ts
import { can, defineShare, objectRef } from "@sixb/core"
import { approveProposal } from "../actions/approve-proposal"
import { LineItem } from "../ontology/line-item"
import { Proposal } from "../ontology/proposal"

export const ProposalApproval = defineShare("proposal-approval", {
  target: Proposal,
  grants: ({ target }) => [
    can.view(target).withLinks([
      Proposal.l.customer,
      Proposal.l.items.withLinks([LineItem.l.product]),
    ]),
    can.apply(approveProposal).on(target),
  ],
})
```

`target` is a contextual placeholder for the exact `ObjectRef` supplied at issuance:

- `can.view(target)` exposes that object only, never every object of its Object Type.
- `.withLinks([Proposal.l.items])` also exposes that edge, its link properties, and its live target
  objects.
- Nested links remain closed unless selected explicitly.
- `.withLinks()` selects every direct link known at issuance, never recursive links.
- `can.apply(action).on(target)` authorizes that exact `(Action, ObjectRef)` pair.

The shared destination is an ordinary app route:

```text
app/proposals/[id]/page.tsx

/proposals/proposal-123                         authenticated entry
/shared/<grantId>/proposals/proposal-123#secret shared entry, same page
```

The page does not use an `app/shared` convention, a shared resource facade, or
`useSharedAccess()`. Its existing standard client and React Query code run unchanged. Shared
bootstrap and query-cache isolation are framework concerns.

Applications should render bearer URLs as native document links. Entering or leaving shared access
must cross a document boundary rather than reuse authenticated in-memory state.

## Grant lifecycle

[...]

PR #410, opened 2026-08-20
core: add shared access grant foundation

quentinnippert:
## Why

Shared links need a framework-owned grant before adding anonymous exchange, sessions, or UI.

This PR establishes that foundation: applications declare the maximum authority of a share, while the framework owns authorization, credentials, and persistence.

## What

| Layer | Added |
| --- | --- |
| Definition | `defineShareType()` and `shares/` auto-discovery |
| Authorization | `can.share(ShareType)` **and** canonical access to the exact target |
| Management API | `POST`, `GET`, and `DELETE /api/share-grants` |
| Credentials | 256-bit secret returned once; only its digest is stored |
| Storage | In-memory, SQLite, and PostgreSQL implementations |
| Client | OpenAPI contract and generated typed client |

```text
authenticated caller
  -> can.share(ShareType)
  -> exact-target access
  -> issue grant
  -> return link once
  -> persist digest only
```

Issuing code cannot widen the share type's grants. V1 accepts exact-object views and object-bound Actions; broad grants and Actions receiving an `objectRef` are rejected.

## Scope

This is the first slice of #269. It intentionally stops before:

- link exchange and short-lived shared sessions;
- shared resource and Action endpoints;
- the isolated shared application entry point and client.

## Validation

- `bun run test:ci` — 3,807 tests passed
- `bun run typecheck`
- `bun run build`
- `bun run check`
- `bun run test:publish`

PostgreSQL e2e coverage is included and runs when `DATABASE_URL` is available.


quentinnippert:
Superseded by the rebuilt #269 stack, primarily #480 and #481. The new model scopes canonical reads to exact resources and expresses sharing through grants instead of the previous foundation. Closing in favor of #480–#483.

PR #413, opened 2026-08-20
core: add shared access sessions

quentinnippert:
## Why

An `anyoneWithLink` grant should not become a normal application session or expose its long-lived
fragment secret on every API call. This slice adds an isolated shared-access session boundary on
top of the grant foundation from #410.

## Flow

```text
/shared/:grantId#secret
          │ POST /api/shares/:grantId/exchange
          ▼
short-lived HttpOnly shared session
          │ SharedAccessGuard + dedicated CSRF
          ▼
shared routes only
```

- Ambient OIDC and bearer authority is ignored.
- Shared cookies are rejected by normal Object and Action routes.
- The grant is rechecked on every request, so expiry or revocation takes effect immediately.
- Link and session secrets are persisted only as digests.

## What this adds

| Layer | Change |
| --- | --- |
| Core | Link exchange, isolated shared principal and 15-minute session lifecycle capped by grant expiry |
| Server | Dedicated guard and `exchange`, `session`, and `sign-out` routes |
| Client | Isolated `@sixb/client/shared` entrypoint with its own CSRF state |
| Storage | Shared-session contract and InMemory, SQLite, and PostgreSQL implementations |

The storage contract suite locks provider parity, defensive copies, project isolation, validation,
collisions, and concurrent idempotent revocation.

This also tightens two details from the grant foundation: HTTP revocation accepts `system`
attribution, and InMemory validation now runs before collision detection like the SQL providers.

## Scope

This PR intentionally stops before the shared page and shared Object/Action execution. Those can
build on this boundary without widening normal application authority.

Part of #269. Stacked on #410.

## Validation

- `bun run test:ci` — 3,843 passed, 0 failed
- `bun run typecheck`
- `bun run build`
- `bun run check`
- `bun run generate:client`
- `bun run test:publish` — 49 packages verified


quentinnippert:
Superseded by #482 in the rebuilt #269 stack. Shared sessions now establish delegated authority for the standard Object, Query, Link, and Action APIs, without a parallel resource API. Closing in favor of #480–#483.

Issue #384, opened 2026-08-14
Idea: connect external accounts from Sixb apps


quentinnippert:
# Idea: connect external accounts from Sixb apps

## Problem

Static connectors work for server-owned credentials, but applications otherwise have to rebuild
OAuth, encrypted token storage, refresh, account selection, and multi-account ingestion.

## V1 model

```text
Connector definition (static)
  └─ Authorization (private encrypted OAuth grant)
       ├─ Connection → account A → slot "organic"
       └─ Connection → account B → slot "paid"
```

- A connection is `project + connector + slot → external account`.
- Project ownership is implicit in V1.
- One slot has at most one connected account; replacement is explicit.
- Several connections may share one authorization without duplicating credentials.
- A durable connection run tracks the headless OAuth/selection flow and contains no secret.

## Developer experience

```ts
// connectors/tiktok.ts
export const tiktok = defineConnector(
  "tiktok",
  tiktokAccounts({ clientId, clientSecret }),
)
```

Durable OAuth storage requires one deployment secret:

```ts
createSixb({
  storage,
  connectorConnections: {
    encryptionKey: process.env.SIXB_CONNECTOR_ENCRYPTION_KEY!,
  },
})
```

Applications own their UI through a headless hook:

```tsx
const connection = useConnectorConnection({
  connectorId: "tiktok",
  slot: "organic",
})

<button onClick={connection.connect} disabled={!connection.canConnect}>
  {connection.connection?.account.label ?? "Connect TikTok"}
</button>
```

OAuth Syncs automatically read every connected account:

```ts
defineSync("sync-tiktok-videos")
  .from(tiktok)
  .read((client, { connection }) => client.listVideos())
  .intoDataset(videos)
```

```text
one Sync run
  ├─ connection A → read
  ├─ connection B → read
  └─ one atomic dataset commit
```

## Contracts

- Sixb owns state, PKCE, callback validation, encrypted persistence, refresh coordination, and
  revocation recovery.
- Management requires CSRF protection and `can.manage(connector)`.
- Tokens, codes, verifiers, client secrets, and authorization ids never reach browser responses.
- Managed clients are available only inside authorized framework executions.
- Sync fan-out is sequential and deterministic; one source failure preserves the previous dataset
  and every checkpoint.
- Static connectors remain backward compatible.

## Implementation stack

1. [#411](https://github.com/sixb-ai/sixb/pull/411) — Core lifecycle, storage contract, encryption,
[...]

PR #414, opened 2026-08-21
storage: add a connector connection store and sealed credentials

yashksaini-coder:
## Summary

The storage foundation for #384, and nothing above it. `Storage.connectorConnections`
holds three sub-stores — authorization attempts, authorizations, and connections —
plus the sealed-credential primitive they depend on, an in-memory implementation,
and a reusable contract suite so `@sixb/pg` and `@sixb/sqlite` can implement against
a fixed shape later.

Attempts are short-lived and single-use, and carry the owner and slot they were
minted for, so a callback that supplies only `state` cannot retarget them.
Authorizations are separate from connections because one grant may expose several
external accounts; those connections share the grant instead of duplicating token
material. Refresh is coordinated per authorization with a lease and a revision:
only `commitRefresh` moves the revision, and it persists rotated credentials and
releases the lease together. A lost compare-and-set throws rather than returning an
ignorable outcome — overwriting a newer refresh token with an older one costs the
user a reconnect.

Credentials are sealed with AES-256-GCM from `node:crypto`, no new dependency.
This is the first reversible secret Sixb persists; every other secret is hashed,
which is not available when the value has to be recovered to call an API. The AEAD
additional data is reconstructed at open time from the row's immutable identity
rather than stored, so ciphertext copied to another project, connector, record, or
purpose stops opening. Storage never holds the key — providers persist an opaque
envelope, which is also what keeps the contract suite key-free.

## Linked issue

Advances #384.

## Scope

Deliberately not here, and deferred to later PRs in the split I proposed on the
issue: OAuth routes, the connect UI, `sixb.connector(..., { owner, slot })`, the
`createSixb({ connectorConnections: { encryptionKey } })` config (nothing reads it
yet), and the Pg/SQLite implementations and their migrations (`020` was taken by the usage-projection
drop while this was in flight, so they will land as `021`).

Also cut for want of a caller: `attempts.getById` (a read that does not consume is
a replay window on a single-use token), `cleanupExpired`, `authorizations.list`,
and `connections.getById`/`delete`. Root cross-aggregate operations
(`completeAuthorization`, `revokeAuthorization`) are left out too — callers compose
inside `storage.transaction()` until the OAuth PR shows what they actually need.

One incidental move: `principalKey` goes from the file-upload-session helpers to
`auth`, beside `principalsEqual`. The owner key needs it, and copying the one line
is how a second, subtly divergent derivation of `${type}:${id}` gets born.

## Validation

`bun run check`, `bun run typecheck`, `bun run build`, `bun run test:publish`, and
`bun run generate:client` (no contract change). Unit suite green apart from four
`create-sixb`/`sixb init` failures that reproduce identically on a clean `main`
under Bun 1.4.0, and two environment-flaky sandbox/sftp tests that pass in
isolation — none of them near this change.

Each guard was proven by reverting it:

- drop `recordId` from the AAD and only the rebinding test fails;
- swap `createConnectorConnectionOperationScope` for the generic facade and only
  the transaction-escape test fails.

That second check corrected a wrong assumption of mine worth stating: the bespoke
scope wrapper does **not** protect rollback — rollback runs off the raw stores, so
it passes either way. What the wrapper prevents is a root-storage write inside a
transaction callback silently landing outside the transaction, because
`ConnectorConnectionStorage` exposes no methods of its own for the generic facade
to wrap. The test now says so.

## Notes

Four decisions the issue does not settle. Each is cheap to change now and
expensive after a migration lands, since migration files are checksum-frozen and
SQLite cannot widen a `CHECK` without a table rebuild:

1. **Connection status is derived, not stored.** The issue says a terminal refresh
   failure changes connection status, but the thing that failed is the
   authorization and one grant may back many connections. Deriving makes that rule
   cost zero writes and makes a stranded connection fail closed by construction.
   Storing it would make it an N-row fan-out with a crash window. This is the one
   choice that changes the Pg schema.
2. **`Date`, not ISO strings.** Matches `AuthOidcAuthorizationAttemptStore`, the
   closest sibling, and this store compares three separate expiries where lexical
   ISO comparison breaks on `+00:00` versus `Z`. Newer stores such as
   `webhook-deliveries` use strings, but they only record timestamps.
3. **How does a second account attach to an existing grant?** Every affordance in
   the issue is one button per `(owner, slot)`, so nothing reaches the shared-grant
   state that lifecycle rules 3-5 describe. I left the store able to express it
   (`listByAuthorization`) without picking the mechanism.
4. **`authorizedBy` is `AuthorizablePrincipal`, not `Principal`** — a `system`
   principal cannot complete an interactive consent. Narrowing later would be
   breaking; widening would not.

Separately, and not blocking this PR: the issue's authorization section cannot be
written as specified today. There is no `can.manage` builder and no connector grant
kind, and all object authorization is type-level, so "the owner's applicable access
contract" cannot distinguish one object instance from another. Happy to open a
separate issue for that if useful.


quentinnippert:
@yashksaini-coder I already opened this PR: https://github.com/sixb-ai/sixb/pull/411 We're probably going to have conflicts.
Do you want to continue the implementation of the issue https://github.com/sixb-ai/sixb/issues/384 from my PR ? I'm down for whatever works the best for you

Thanks again for contributing, we appreciate that a lot.

Could be easier to talk on discord if you want to join: https://discord.gg/z88w94Muy

yashksaini-coder:
@quentinnippert  Apology, I didn't see the PR attached, Github has been acting strange, it wasn't showing the pr. Glad I made it a draft, closing it now

quentinnippert:
> @quentinnippert Apology, I didn't see the PR attached, Github has been acting strange, it wasn't showing the pr. Glad I made it a draft, closing it now

@yashksaini-coder no worries, that's on us. We're still figuring out how to make it easy to contribute

PR #418, opened 2026-08-23
scheduler: keep long delays and unschedulable expressions from stopping the scheduler

yashksaini-coder:
## Summary

Two independent ways the scheduler could stop doing its job. Both are in
`SchedulerRuntime`; neither is reachable through a config error the validator catches.

**`armTimer` passed a raw delay to `setTimeout`.** Node coerces anything above 2147483647ms to
`1`, so any occurrence more than ~24.8 days out — every monthly and yearly cron — armed a timer
that fired immediately, found nothing due, re-armed, and repeated. That spins the event loop and
emits a `TimeoutOverflowWarning` per re-arm for the whole wait: roughly five days for a monthly
schedule, eleven months for a yearly one. `nextCronOccurrence` searches up to four years ahead, so
the returned instant can be ~58x over the ceiling.

Clamping is safe because `tick()` re-arms unconditionally from the same occurrence map: a clamped
delay just becomes several sleeps instead of one, and an early wake finds nothing due and does
nothing. Roughly fifteen extra wakes a year for a yearly cron.

**`start()` called `nextCronOccurrence` unguarded.** A syntactically valid expression can still
have no reachable occurrence — `0 0 30 2 *` parses, passes `validateSchedulesAtStartup`, and then
throws "No matching occurrence found within 4 years". That escaped `start()`, which
`runScheduler` turns into `process.exit(1)`, so one typo stopped every other schedule in the
project and crash-looped the scheduler role.

`tick()` already tolerates exactly this error per schedule; start-up now agrees with it. Making
start-up and steady state disagree on the same error seemed worse than either policy alone.

## Linked issue

None — found while reading the scheduler.

## Scope

`packages/core/src/scheduler/runtime.ts` and its test file. Deliberately not included:

- Making an unreachable expression a **startup** error. Arguably the better home for loudness is
  `validateSchedulesAtStartup`, which already calls `createCronMatcher` — but a reachability probe
  there would newly reject leap-day expressions sitting on the four-year boundary, so it wants its
  own discussion.
- Persisting a last-fired watermark. Occurrences during downtime are dropped today; that is a
  design question, not a bug.

While in here I also noticed two DST issues in `next-occurrence.ts` (a spring-forward skip and a
fall-back double-fire for timezone'd schedules) and that `N/step` silently drops the step. Happy to
open those separately — they seemed too unrelated to bundle.

## Validation

`bun run check`, `bun run typecheck`, `bun run build`, `bun run test:publish`, and the full unit
suite (3884 pass; the only failures are four `create-sixb`/`sixb init` tests that reproduce
identically on a clean `main` under Bun 1.4.0). `generate:client` is unaffected — no routes or
schemas change.

Both fixes are covered by tests that fail when their guard is removed:

- clamp reverted → the timer records 1001 wakes in one simulated second instead of 1
- `try/catch` reverted → `CronValidationError` escapes `start()`

### DIFF
--- a/packages/core/src/scheduler/runtime.ts
+++ b/packages/core/src/scheduler/runtime.ts
@@ -6,6 +6,15 @@ import { SchedulerValidationError } from "./errors"
 /** Matches the `onError` flush budget: long enough for a healthy broker, short enough to shut down. */
 const DEFAULT_EMIT_DRAIN_TIMEOUT_MS = 5_000
 
+/**
+ * Largest delay `setTimeout` accepts before it overflows its 32-bit counter.
+ *
+ * Node coerces anything larger to `1`, so an unclamped monthly or yearly occurrence would fire
+ * immediately, find nothing due, and re-arm on the next tick -- spinning the event loop and
+ * emitting a `TimeoutOverflowWarning` for every re-arm until the occurrence came into range.
+ */
+const MAX_TIMER_DELAY_MS = 2_147_483_647
+
 export interface SchedulerRuntimeOptions {
   schedules: readonly ScheduleDefinition[]
   events: DomainEventLog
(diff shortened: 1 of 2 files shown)


quentinnippert (approved):

quentinnippert:
@yashksaini-coder love that, great finding. thank you!

PR #419, opened 2026-08-23
brokers: reject a subscribe cursor that predates the retained range

yashksaini-coder:
## Summary

`read` and `tail` both refuse an `afterCursor` older than what a stream still holds, because
silently advancing to the oldest retained record hands a caller a gap it cannot detect. `subscribe`
did not, in two of the three providers.

A subscriber resuming from a trimmed cursor was fast-forwarded to whatever survived retention and
delivered that as the continuation. Nothing reported the gap, so the caller saw a complete stream.

This is reachable from a client, not just in theory: `/ws/agents` passes a caller-supplied
`afterCursor` straight into `subscribe`, and agent run streams carry retention.

Redis already got this right, which is what made the divergence easy to miss. The in-memory
provider now checks before registering the subscription, so a rejected resume cannot leave one
behind; JetStream checks before building the consumer, and only when `afterCursor` is set, so
`from: "latest"` and `from: "earliest"` are unchanged.

**The second commit is the part I would not ship without.** Rejecting the cursor on its own turns a
silent gap into a dead socket: `/ws/agents` caught the error, dropped the subscription and sent
`{type:"error"}`, and the client handler only records the error — it does not close the socket,
does not reconnect, and does not clear its stored cursor. The socket stayed open holding a cursor
the broker would reject forever.

That is not a rare state. An agent run stream retains 5000 records and a streamed turn writes one
per chunk, so a long turn plus a backgrounded tab trims the resume point routinely. The websocket
now retries from the oldest retained record and reports `afterCursor: null` on the `subscribed`
frame, so the client learns its resume point is gone and replays what survives. `log-subscription-hub`
already does the equivalent with its `cursor_expired` reset.

## Linked issue

None — found while reading the broker providers.

## Scope

The contract suite covered `read` and `tail` but not `subscribe`, which is how two providers
drifted. It now covers all three, so every provider is held to the same rule — including a
third-party one. The rule itself also moved onto `subscribe`'s own contract docs; it lived only on
`read`'s, which is the other half of why this was easy to miss.

Not included: the pre-existing JetStream race where a stream can be trimmed between the info fetch
and consumer creation. `read` has the identical exposure and closing it needs an ack-time sequence
check.

## Validation

`bun run check`, `bun run typecheck`, `bun run build`, `bun run test:publish`, `generate:client`
(unchanged), and the full unit suite — 3884 pass, with only the four `create-sixb`/`sixb init`
failures that reproduce on a clean `main` under Bun 1.4.0.

Because this changes a shared contract, I ran the suite against real servers rather than in-memory
alone: **NATS 29/29** and **Redis 42/42** via each package's `docker-compose.yml`. The added case
fails against either changed provider with its guard removed, and the in-memory guard-removal check
is noted in the test.

I also traced every `broker.subscribe` call site for callers that pass `afterCursor`. `/ws/agents`
is the only non-test one; `ws/events`, `log-subscription-hub`, the orchestrator and the outbox
dispatcher all subscribe without a cursor and are unaffected.

### DIFF
--- a/broker/nats/src/nats-broker.ts
+++ b/broker/nats/src/nats-broker.ts
@@ -310,6 +310,15 @@ export class NatsBroker implements Broker {
 
     let stopped = false
     const streamName = await this.requireStreamName(params.projectId, params.streamId)
+
+    // A cursor JetStream has already discarded would silently start the consumer at the oldest
+    // retained message, so a resuming subscriber would skip the gap without noticing. `read` and
+    // `tail` both reject it; this keeps `subscribe` consistent with them.
+    if (params.afterCursor !== undefined) {
+      const streamInfo = await this.fetchStreamInfo(streamName)
+      this.assertCursorInRetainedRange(streamInfo, params.streamId, params.afterCursor)
+    }
+
     const js = await this.getJetStream()
     const consumerOptions = this.consumerOptions({
       projectId: params.projectId,
--- a/docs/agents/running-and-streaming.md
+++ b/docs/agents/running-and-streaming.md
@@ -127,7 +127,10 @@ Each `record` frame carries an `AgentRunStreamEvent`:
 | `agent.run.finished` | The run ended. | `status`, `finishReason`, `error` |
 
 Records carry a **cursor**. Persist the last one you saw and pass it as `afterCursor` to resume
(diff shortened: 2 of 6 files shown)


quentinnippert (approved):
you're cooking

Issue #290, opened 2026-08-02
Idea: control agent token usage by project, group, and principal


demattosanthony:
## Summary

Add provider-neutral execution limits to Sixb agents and DB-backed monthly token quotas for
organizations.

The split is intentional:

```text
agent definition                    database
────────────────                    ────────
max output tokens per model call    project monthly quota
max total tokens per run            shared group monthly quota
max steps                           user/service-account monthly quota
```

Agent limits belong in code because they define how an agent runs. Organization quotas belong in
storage because operators need to change them without editing or redeploying a project.

V1 is an **enforced token quota with bounded overrun**, not an exact spend cap. A model call already
in progress may finish over the limit, and concurrent calls may cross it before their usage is
recorded.

## Problem

Sixb records provider-reported token usage after agent runs, but cannot use it to control future
usage.

Today there is no way to:

- cap one agent run by tokens;
- set a shared project or group quota;
- limit one user or service account;
- account consistently across conversations and workflow agent nodes;
- stop another model call after a quota is exhausted.

Direct runs also do not snapshot requester groups, workflow agent usage is stored separately, and
failed or cancelled runs may lose usage returned before the failure.

### Bound individual agents in code

Extend `defineAgent` with a per-call output limit and a total-run token stop condition:

```ts
export const supportAgent = defineAgent("support", {
  name: "Support Agent",
  model: gateway("openai/gpt-5"),
  instructions: "Help the support team resolve customer issues.",
  maxOutputTokens: 8_000,
  loop: {
    stopWhen: {
      maxSteps: 25,
      maxTokens: 100_000,
    },
  },
})
```

`maxOutputTokens` limits one provider call. `maxTokens` stops the loop after completed model calls
reach the run limit. The final call may cross the total, but these controls prevent an unbounded tool
loop and limit aggregate-quota overrun.

### Store organization quotas in the database

```ts
interface TokenQuota {
  subject:
    | { type: "project" }
    | { type: "group"; id: string }
    | { type: "user"; id: string }
    | { type: "serviceAccount"; id: string }
  limitTokens: number
  period: "month"
  enabled: boolean
}
```

Example:

```json
{
  "subject": { "type": "group", "id": "support" },
  "limitTokens": 5000000,
  "period": "month",
  "enabled": true
}
```

V1 uses fixed calendar months in UTC and allows one active monthly quota per subject.

```text
project quota      shared by every agent execution in the project
group quota        shared by all attributed members of that group
[...]

PR #431, opened 2026-08-25
core,agent-worker: add a framework-managed main agent with sub_agent

yashksaini-coder:
# core,agent-worker: add a framework-managed main agent with sub_agent

Draft for #290, implementing the shape you described. Runtime half only — no UI, see **Scope**.

## What this adds

```ts
createSixb({
  mainAgent: { name: "Assistant", model, instructions: "..." },
})
```

Registered under the reserved id `main`, framework-managed, no `tools` and no `groups`. It receives
one injected tool, `sub_agent`, which can invoke any project agent the *requester* is allowed to run.

A delegated run is a real run: its own `agent_run`, its own execution linked through
`source.executionId`, its own service account, sandbox and model-call ledger rows, on its own thread.

## Three places this departs from your reply

**1. The child runs in the parent's worker slot, not through the queue.**

You flagged that waiting main agents must not occupy every slot their children need. Moving the
default 4 → 8 does not fix that — it relocates it to 8 blocked parents, then wedges until eight
5-minute turn timeouts fire, with the 60s lease renewing throughout so nothing rescues it.

The workflow agent node dodges this by *parking* the parent (`agent-node-executor.ts:105-173`), but
that cannot transfer: a workflow node is a durable state machine, an agent turn is a live
`streamText` loop, and you cannot park mid-tool-call without durable continuations.

So `sub_agent` admits the child and runs its turn inline via the same `runAgentTurn`. The
concurrency default stays 4. Everything you asked for is preserved.

One consequence worth naming: the child is created **and started in one transaction**, because
`dispatchQueuedAgentRuns` scans every run in `queued` with no filter (`worker.ts:444-448`). A child
observably `queued` would be claimed and either started twice or reclaimed out from under the
in-process turn. Pleasingly, that also gives us crash recovery for free — a parent that dies
*before* the child starts leaves it `queued`, and the same scan picks it up.

**2. A new `requesterAuthorizationGroupIds` column on `agent_runs`.**

The main agent runs with its requester's authority, as you asked. The obvious way to rebuild that
authority in the worker — `run.requesterGroupIds` — is a privilege escalation.
`snapshotRequesterGroupIds` says so in its own doc comment (`auth/attribution.ts:5-10`): it is
deliberately the principal's *full* membership, because accounting must not inherit a token's
constraint. `constrainTokenGroupIds` (`auth/runtime.ts:1503-1513`) narrows only the authorization
context.

Concretely: a user in `{support, admin}` mints a token scoped to `["support"]`.
`POST /api/agents/db-admin` is refused, but talking to the main agent would reach it.

So admission persists the already-constrained set (`runtime.authorization?.groupIds`) and both the
run's own reach and its delegation decisions resolve from that, with roles still live.
`requesterGroupIds` is untouched for accounting. Migration `025` on both providers.

**3. No UI.**

#420 rewrites `AgentChat.tsx` and `ConversationPanel.tsx` and keeps the picker, so I have left the
UI alone entirely. It should need very little later: `canGoHome` is already
`!pinnedAgentId && agents.length > 1` and `AgentIdentity` self-hides at `agents.length <= 1`. Child
threads are owned by `svc_agent_main` and thread listing is owner-filtered, so they are already
invisible with no code change.

## Authorization shape

```text
User request
  └── main — runs as the requesting user, no groups of its own
        └── sub_agent(target)  ← checks the requester's run:agent(target)
              └── target — own service account, own groups, own sandbox
```

**The main agent runs with the initiating user's authority**, so two people talking to the same
assistant each reach exactly their own data from one definition. That is why `mainAgent` takes no
`groups`: a fixed identity would flatten every user's reach into one. Delegation does not pass the
identity onward — the child still runs under the target agent's own service account and groups, as
you specified.

The rule the contract now enforces is that an Agent execution acts under **either** its own
`svc_agent_<id>` **or** exactly its own `requestedBy`. That is self-describing, so it needs no extra
column to distinguish the two cases. Six enforcement points had to widen, not the five I first
counted — `execution/authorization.ts:250` validates the agent execution *binding* on every
`withScope`, and only surfaced when a delegated turn silently failed to start:

```text
execution/agent.ts               createAgentExecutionRecord / assertAgentExecutionRecord
execution/authorization.ts:59    createAgentRuntimeAuthorization
execution/authorization.ts:250   assertCompatibleExecutionAuthority   ← the one I missed
agents/authority.ts              resolveAgentRunAuthorization (new dispatcher)
storage/executions/run-link.ts   findAgentRunExecution
storage/executions/validation.ts provider rules + both SQL CHECK constraints
```

A main-agent turn with no human behind it (schedule, system trigger) falls back to its own service
account, which has no groups: unattended work does not silently inherit anyone's authority.

**The consequence worth reviewing carefully:** the main agent's sandbox reaches the Sixb API through
its run-scoped gateway, so model-written `bash` commands now execute with the *user's* permissions
rather than a fixed service account's. That is precisely what "runs with the initiating user's
authority" means, and the gateway's 29-route allow-list still bounds which endpoints exist at all —
but it is the line to disagree with if you meant something narrower.

Authorship is deliberately **not** delegated. Assistant messages are stamped with
`svc_agent_<agentId>`, not the run's authority; recording the human as the author of the agent's own
replies would misattribute the transcript, and `authorPrincipal` is exposed over HTTP. I had this
wrong at first — `run-agent-turn.ts` used `context.agentPrincipal`, which the change had quietly
turned into the user. Same class of mistake as the child-thread owner, and both now have tests that
fail if reverted.

### Blast radius

Checked and unaffected: `serializeAgentRun` is an explicit allow-list, so neither requester-group
field is exposed over HTTP; the workflow agent-node path still resolves service-account authority
only and fails closed on anything else; `bindDurableAgentExecution` has a single consumer; no UI
reads `authorPrincipal`/`ownerPrincipal`/`requestedBy`. The SQLite rebuild drops no indexes, views
or triggers (`executions` has none) and its column list is byte-identical to the one `022` created,
so the `SELECT *` copy cannot misalign — there is a migration test that seeds a row, migrates, and
asserts both that it survives and that the delegated form flips from rejected to accepted.

Only the main agent is given `sub_agent`, and the core entry point refuses `main` as a target, so
delegation depth is bounded at one structurally rather than by a counter.

## Known limits, all documented in `docs/agents/main-agent.md`

- Delegation is **at-least-once**: a redelivered main-agent turn replays and may re-delegate.
- A child has no queue job, so one orphaned *after* start stays `running`. Inert today (single-use
  thread, owned by `svc_agent_main`, listed nowhere) — but a startup sweep is a hard prerequisite for
  the sub-agent-thread UI work, not an optional follow-up.
- No parent-side usage rollup: `ExecutionStorage` exposes only `create` and `getById`, so walking
  `source.executionId` needs a new query.
- `sub_agent` calls are serialized, because every execution provisions a sandbox eagerly.
- The deadlock fix covers `sub_agent` only. `requestWorkflowRun` is `agentApi: true` and a workflow
  `agentStep` still enqueues onto the same lane — pre-existing, and its own PR.

## One thing that needs your call

**`run:agent` on the main agent alone is not expressible.** `can.run` takes a definition object and
`createMainAgentDefinition` is framework-owned, so the only grant that covers `main` is
`can.run(every.agent())` — which you did mention, but it means "let this group talk to the assistant,
nothing else" cannot be written. Happy to add a selector if you want one.

## Validation

- `bun run check`
- `bun run typecheck`
- `bun run build && bun run test:publish` — 49 packages
- `bun run generate:client` — no drift, this PR touches no HTTP contract
- `bun run test:ci`
- `bun --filter @sixb/agent-worker test:e2e` — new lane, see below

Four failures in `create-sixb` / `sixb init` reproduce on a clean checkout of `main` and are
unrelated.

Every guard was proven by reverting the fix and watching it fail:

| Guard | Reverted to | Observed |
| --- | --- | --- |
| reserved-id message | drop the explicit check | throws `Duplicate agent id: main` instead |
| `main` reaches `SecurityRegistry` | index only declared agents | `every.agent()` resolves without `main` |
| null requester denies | fold into `isAllowed(...)` | delegation admitted for an unauthenticated run |
| constrained groups | authorize off `requesterGroupIds` | reaching `db-admin` **succeeds** |
| child never queued | dispatch the child | job id appears on the agent queue |
| never left `queued` | return before `runs.start` | run comes back `queued` |
| swallowed child finalize | drop `assertToolsHealthy` | parent acks on attempt 1 instead of redelivering |

That last one is why `packages/agent-worker` gains a `test:e2e` script: the delegating job waits out
`FINALIZE_RETRY_BACKOFF_MS` (5s) before redelivering, which is longer than the unit suite should ever
go quiet.


quentinnippert:
Thank you for the PR, we appreciate it a lot. We decided to go in this direction: https://github.com/sixb-ai/sixb/issues/451
Could be good to exchange more before implementing, let me know if you think about a better process.

PR #477, opened 2026-08-31
agents: POC own the LLM inference runtime

demattosanthony:
## Summary

Sixb owns language-model inference through a provider-neutral contract in `@sixb/core`.

```text
@sixb/core/models
  model contract
  messages and stream events
  tools and structured output
  model definitions and catalogs
  reasoning capabilities
  exact cost estimates and provider-reported costs

@sixb/core agent runtime
  multi-step model loop
  tool execution
  structured-output validation
  provider-state replay

@sixb/agent-worker
  agent and workflow adapters
  context budgeting and compaction
  durable usage and cost accounting

models/
  anthropic/
  vercel-ai-gateway/
```

## Authoring

```ts
// agents/operations-assistant.ts
import { defineAgent } from "@sixb/core"
import { vercelGateway } from "@sixb/vercel-ai-gateway"

export const operationsAssistant = defineAgent("operations-assistant", {
  name: "Operations Assistant",
  model: vercelGateway("openai/gpt-5.5"),
  reasoning: "medium",
  instructions: "Help operate this project using authorized data and tools.",
})
```

Direct providers use the same callable shape.

```ts
import { anthropic } from "@sixb/anthropic"

export const supportAgent = defineAgent("support-agent", {
  name: "Support Agent",
  model: anthropic("claude-sonnet-5"),
  reasoning: "high",
  instructions: "Help customers using verified account and product information.",
})
```

Model IDs remain strings so new provider models work without a Sixb release.

```ts
vercelGateway("openai/gpt-5.5")
vercelGateway("anthropic/claude-sonnet-4.6")
anthropic("claude-sonnet-5")
```

## Project model catalog

```ts
import { createSixb } from "@sixb/core"
import { anthropic } from "@sixb/anthropic"
import { vercelGateway } from "@sixb/vercel-ai-gateway"

export const sixb = createSixb({
  id: "acme",
  models: {
    language: [
      vercelGateway("openai/gpt-5.5"),
      vercelGateway("anthropic/claude-sonnet-4.6"),
      anthropic("claude-sonnet-5"),
    ],
  },
})
```

```ts
sixb.definitions.models?.language.default
sixb.definitions.models?.language.getById("vercel-ai-gateway/openai/gpt-5.5")
sixb.definitions.models?.language.list()
```

The catalog is optional. When configured, it must be nonempty and every agent's provider/model pair
must appear in it. The first model is the default, and duplicate bindings fail at startup. The same
vendor model through a gateway and a direct provider are separate entries because they route and
[...]

Issue #451, opened 2026-08-26
Idea: make one autonomous agent the project entry point


quentinnippert:
> Replace the static agent catalog with one built-in main agent that can delegate work to short-lived agents using project-approved models.

## Why

Today, developers must design the agent roster in advance:

| Today | Proposed |
| --- | --- |
| Define each agent with `defineAgent` | Configure the models Sixb may use |
| User chooses which agent to contact | User always talks to the main agent |
| Model and instructions are fixed | Main agent configures each child for its task |
| Workflow steps reference an agent definition | Workflow steps configure their task directly |

This makes delegation static and exposes framework structure to users. The project should define its available capabilities, not a fixed organization of agents.

## Proposed DX

```ts
import { gateway } from "ai"

export const sixb = createSixb({
  // ...

  models: {
    language: [
      gateway("openai/gpt-5.4"),
      gateway("anthropic/claude-sonnet-4.6"),
    ],
  },
})
```

These are regular AI SDK models. Direct provider models remain supported.

- The first language model is the default.
- Sixb derives internal references from `provider + modelId`.
- Developers do not define additional IDs or aliases.
- The UI exposes the configured models and supported reasoning levels.
- There is no `mainAgent`, `agentLimits`, custom router, or agent picker configuration.

The catalog is organized by technical model kind so it can grow without changing shape:

```ts
models: {
  language: [...],       // v1
  image: [...],          // later
  speech: [...],         // later
  transcription: [...],  // later
}
```

## Agent experience

```text
User
 └─ Main agent
     ├─ Child agent: focused instructions, fast model
     └─ Child agent: focused instructions, deeper reasoning
```

The framework owns the main agent and gives it a `spawn_agent` tool.

A child receives a task, a model from `models.language`, a reasoning level, and specific instructions. It has its own run, execution, stream, and sandbox. Its result returns to the parent; it does not create a user-visible thread.

In v1:

- only the main agent can spawn children;
- children cannot spawn more children;
- fan-out is bounded internally, without public limit configuration.

The interface contains one chat experience for the main agent. Child activity may be shown within that conversation, but users do not chat with children directly.

## Workflow agent tasks

Workflow steps should configure the task instead of referencing `defineAgent`:

```ts
const extractOrder = defineAgentStep("extract-order", {
  model: gateway("openai/gpt-5.4"),
  reasoning: "medium",
  groups: [orderAutomation],
  instructions: "Only extract information supported by the source.",
})
  .input({ /* ... */ })
[...]

PR #478, opened 2026-08-31
core,server: add the project model catalog

yashksaini-coder:
First slice of #451. @quentinnippert you said you're on #451 — this is one piece you can take,
leave, or redirect.

## Why

#451 replaces the static agent catalog with a project-level model catalog, one framework-owned main
agent, and inline workflow task configuration. The model catalog is the piece every other part
consumes — the main agent picks a child's model from it, and a workflow agent step's model "must
belong to the project catalog" — so it lands first and alone.

This slice is deliberately **additive**. `defineAgent` is untouched, nothing is removed, and
`models` is optional: a project that does not configure it behaves exactly as before. The main
agent, `spawn_agent`, run kinds, the `defineAgentStep` reshaping, and removing the agent catalog and
picker are all later slices.

## What changes

- `createSixb({ models: { language: [...] } })` — an ordered list of AI SDK models, first entry is
  the project default. Keyed by technical model kind so `image`, `speech`, and `transcription` can
  be added later without changing the shape.
- Sixb derives an internal reference per entry; projects never author an id or an alias.
- The catalog rejects two models that derive the same reference, and an empty `language` list, at
  construction.
- When a catalog is configured, `createSixb()` fails at startup if an agent uses a model outside it.
- `GET /api/models` returns the configured models with the default flagged, as a sibling of
  `GET /api/agents` (session-scoped the same way; neither is in `SIXB_API_ROUTES`).

## Two decisions worth your eye

**An entry is a binding, not a vendor model.** `gateway("openai/gpt-5.4")` and `openai("gpt-5.4")`
are two entries, because they route and bill differently. That follows from the AI SDK's `provider`
naming the binding — `openai.chat` and `openai.responses` are distinct — which is also why the
internal reference stays internal and is **not** in the API response: it is not the canonical id,
and `SDK_PROVIDER_BINDINGS` in `@sixb/agent-worker` is what resolves that for billing. The response
carries the `provider` and `modelId` the project configured, which together identify an entry
without inventing a second identity contract. If you would rather the catalog canonicalize up front,
that map wants to move into `@sixb/core` and I am happy to do it — I left it alone because it is
adjacent to the pricing work that just landed, and because #477 makes `providerId` first-class
anyway.

**Catalog membership is enforced for `defineAgent`, which #451 does not ask for.** The issue
requires it for workflow steps and children; I extended it to agents so a configured catalog is not
inert in this slice. It is startup-fatal, and `defineAgent` is slated for removal, so say the word
and I will drop it.

## On reasoning levels

#451 wants the UI to show "supported reasoning levels", but `AgentReasoningLevel` is a fixed AI SDK
enum and the SDK reports no per-model capabilities. Rather than ship a static capability table that
goes stale, or a constant array repeated on every row that carries no per-entry information, the
route omits the field. #477's `LanguageModel` carries `capabilities`, which would be a real source
of truth; happy to wire it if that lands.

## Verification

`bun run check`, `bun run typecheck`, `bun run test:ci` (4437 pass), `bun run build`,
`bun run test:publish` (52 packages), `bun run generate:client` (no drift). No migration in this
slice, so the pg e2e lane has nothing to reach.

Four tests fail on this branch — `create-sixb packed artifacts` (3) and `sixb init` (1). They fail
identically on a clean `main` in my environment and are unrelated to this change.

Per AGENTS.md all three guards were proven by removal — the duplicate-reference check, the
empty-catalog check, and the agent cross-reference check each fail their own test and only their
own test when reverted. Each test says so, and asserts on the message rather than a bare
`RuntimeError`, which an unrelated startup failure would also satisfy.


quentinnippert (approved):

quentinnippert:
clean, simple, love it

PR #499, opened 2026-09-01
core: bind delegated object queries

quentinnippert:
## Why

- #498 adds path-sensitive Query admission, but delegated readers still reject every Query terminal.
- Shared pages need the normal Query API without route-specific resource loading.

## What

```text
request snapshot
  → ordinary validation
  → selected-path admission
  → canonical executor
  → selected storage
```

- Activates `executeQuery`, `count`, `exists`, `facet`, and `queryLinks`.
- Builds admission and selected storage once from the same compiled scope.
- Executes the exact admitted snapshot; no forgeable executor flag or second AST walker.
- Reuses terminal preflights for facets and physical links.
- Bounds facet fan-out to 16 before reading caller-owned elements.

## Security contract

| Case | Result |
| --- | --- |
| Unselected property or link path | denied before storage |
| Known ID of an unselected sibling | empty result |
| Low-level executor + delegated token | denied |
| Principal / unrestricted reader | unchanged |

## Validation

- Core typecheck
- Biome
- 108 focused tests / 529 assertions
- InMemory end-to-end sibling-ID coverage

## Stack

- Base: #498
- Next: #500

### DIFF
--- a/packages/core/src/agents/context-resolution.ts
+++ b/packages/core/src/agents/context-resolution.ts
@@ -1,4 +1,3 @@
-import { assertAuthorized } from "../authorization"
 import type { SixbRuntimeContext } from "../runtime/types"
 import {
   type AgentContextEntryInput,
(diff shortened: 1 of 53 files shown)


quentinnippert:
Consolidated into #498 for a cohesive review in stack #543. All changes are preserved as focused commits; this PR is superseded, not merged. Its discussion and branch remain available.

PR #529, opened 2026-09-05
agents: retire static agent definitions

quentinnippert:
## Why

Complete the move to one project Agent by removing static definitions and their transitional runtime paths.

Stacked on #528. Part of #451.

## What changes

- Configure capabilities through `createSixb({ models, tools })`; remove `defineAgent`, `agents/` discovery, and the static catalog.
- Use `sixb.agent` and `GET /api/agent`, without an Agent selector. `can.run(agent)` grants the single `run:agent` capability.
- Keep conversation authority inherited from the user and workflow tasks under their managed service accounts.
- Name workflow execution identity `actorId`; expose `workflowId` + `agentStepId` in usage reporting instead of that internal key.
- Remove definition adapters and the redundant conversational instruction; retain Sixb's system rules, workflow instructions, and runtime defaults.

## Upgrade notes

- Breaking API change: move project capabilities to `models` / `tools`, and update callers to `sixb.agent`.
- SQLite and PostgreSQL migrations preserve conversation history while dropping its former Agent identity. Workflow service-account IDs remain unchanged.
- HTTP schemas, generated client, UI, docs, and tests follow the new contracts.

## Verification

- 4,722 tests passed (7 skipped); typecheck, build, and 52-package publish checks passed.
- Client regenerated with no drift.
- Regression coverage for workflow attribution, authorization, and prompt construction.
- PostgreSQL E2E not run locally: Docker is stopped.


quentinnippert:
@demattosanthony the stack is ready for a check. We can probably merge the llm POC PR before and I'll update the PRs of the stack from there

demattosanthony (approved):

PR #541, opened 2026-09-08
actions: expose authority-scoped inert descriptors

quentinnippert:
Part of #269 · Stack #504 · Parent: #506

## What

- Return detached, inert Action descriptors from the execution-bound SDK.
- Expose binding IDs and phase availability, never executable handlers or live ObjectTypes.
- Keep existing principal visibility; delegated discovery requires an exact allowed Action/type pair.
- Apply the selected reader's output budget to delegated metadata.

## Why

Ordinary shared pages need Action metadata to render forms without receiving host-side execution handles.

## Compatibility

- Same HTTP/OpenAPI response shape; no client regeneration required.
- Action dispatch remains separately authorized.
- Shared sessions and durable Action-run access follow in #482.

## Checks

Descriptor detachment, exact delegated admission, server Action routes, typecheck and Biome.

### DIFF
--- a/packages/core/src/actions/delegated-admission.ts
+++ b/packages/core/src/actions/delegated-admission.ts
@@ -1,0 +1,72 @@
+import { AuthorizationError } from "../authorization"
+import type { ResolvedRuntimeAuthorization } from "../execution/authorization"
+import {
+  type AuthorizedObjectReader,
+  assertAuthorizedObjectReaderBinding,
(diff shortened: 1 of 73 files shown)


quentinnippert:
Consolidated into #506 for a cohesive review in stack #543. All changes are preserved as focused commits; this PR is superseded, not merged. Its discussion and branch remain available.

Issue #373, opened 2026-08-12
Idea: give agent threads a persistent workspace


quentinnippert:
# Idea: give agent threads a persistent workspace

## Problem

Every agent run currently gets a fresh sandbox. Files disappear at teardown, so an agent cannot work naturally on the same repository across several messages.

Target: one agent can work on a repository for the lifetime of a thread — edit, run and test it — while Git credentials remain usable but unreadable from the sandbox.

## V1

No new `defineWorkspace` primitive.

```text
defineAgent (recipe) + thread params (binding) → named sandbox (filesystem state)
```

- The agent declares how its sandbox is resolved.
- The thread stores validated, immutable params.
- Sixb derives a stable sandbox name from the project and thread.
- The provider resumes it with `getOrCreate()` and snapshots it with `stop()`.
- A requested capability that the provider cannot guarantee fails closed.

This avoids a workspace registry/table, framework file capture, per-turn cloning and a Sixb egress proxy in V1.

## DX

```ts
export const engineering = defineAgent("engineering", {
  name: "Engineering",
  model,
  instructions: "Work on the client application.",

  params: {
    clientId: "string",
  },

  sandbox: {
    persistence: {
      idleRetention: "7d",
    },

    resolve: async ({ params, sixb }) => {
      const client = await sixb.objects(Client).get(params.clientId)

      return {
        source: git({
          url: client.repositoryUrl,
          revision: client.defaultBranch,
          access: "write",
          auth: githubApp({
            installationId: client.githubInstallationId,
          }),
        }),

        setup: ["bun install"],

        env: {
          // Guest-readable by design: suitable for running the application.
          DATABASE_URL: client.developmentDatabaseUrl,
        },
      }
    },
  },
})

await sixb.agents.threads.create({
  agentId: "engineering",
  params: {
    clientId: "acme",
  },
})
```

`resolve()` runs host-side. Only the params are stored on the thread. Environment variables are explicitly documented as readable by the agent; credentials use a separate delivery path.

One agent can therefore serve any number of repositories without one definition per repository.

## Lifecycle

```text
create/resume → resolve recipe → install auth lease → run → clean run state → stop/snapshot
```

- On first creation, Sixb clones the source and runs `setup`.
- On later turns, the provider restores the same filesystem.
- Sixb verifies a non-secret source fingerprint before running. A mismatch must fail clearly rather than silently use the wrong checkout.
- Run-scoped attachments, capability URLs and credential rules are removed before the snapshot.
[...]

PR #571, opened 2026-09-13
sandboxes: add explicit persistent create and resume

quentinnippert:
## What this adds

First slice of #373: optional named sandbox persistence, implemented by Vercel. Agent runs remain ephemeral; workspace binding and execution follow in the stacked PRs.

```ts
await factory.create() // Ephemeral

if (!factory.resume) throw new Error("Persistent sandboxes are required")

const first = await factory.create({ persistence: { name: "workspace" } })
await first.writeFiles([{ path: "notes.txt", contents: "Work in progress" }])
await first.stop() // Confirms preservation

const next = await factory.resume("workspace")
await next.runCommand("cat", ["notes.txt"])
await next.stop()
// next.destroy() is explicit permanent deletion, not routine teardown.
```

## Guarantees and boundaries

- One creation API; unsupported providers reject persistence before provisioning.
- Create rejects existing names. Resume rejects running, missing or expired state; it never creates a replacement.
- Persistent operations target one SDK session, preventing automatic resume into another VM mid-run.
- `stop()` confirms a successful snapshot; concurrent/repeated callers share its outcome, including failures.
- Run environment is supplied per command; session network access closes before saving. Saved files do not establish authority.
- Retention belongs to provider configuration. Callers must serialize the complete lifecycle; this is not a distributed lock or a durable file store.

## Migration

- Replaces `factory.persistence` with `create({ persistence: { name } })` and optional `resume()`.
- Removes Vercel's legacy `persistent` factory option, which bypassed the new preservation guarantees. Old configuration fails with migration guidance; retention alone does not enable persistence.
- `ResumeSandboxOptions` accepts execution settings only, not creation options.

## Validation

- 5,269 unit tests passed; 7 opt-in skips. 42 targeted tests passed.
- Source/test/template/example typechecks, lint, builds and 56-package publication checks passed.
- New guards were removed to prove regression tests fail, then restored.
- Tests use SDK 2.3.0 with simulated transport: no Vercel usage. Actual service-side preservation remains unverified.


demattosanthony:
Should we keep one creation path, with persistence as an option and an optional `resume()` method?

```ts
interface SandboxFactory {
  create(options?: CreateSandboxOptions): Promise<Sandbox>

  // Present when the provider supports persistent creation and resume.
  resume?(name: string, options?: ResumeSandboxOptions): Promise<Sandbox>
}
```

```ts
// Ephemeral
await factory.create(options)

// Check persistence support
if (!factory.resume) {
  throw new Error("This provider does not support persistent sandboxes")
}

// Create new named state; reject if it already exists
const first = await factory.create({
  ...options,
  persistence: { name: workspaceName },
})
await first.stop() // Confirm preservation

// Resume existing state; reject if missing/expired, never recreate
const next = await factory.resume(workspaceName, options)
await next.stop()

// Explicit permanent deletion
await next.destroy()
```

This avoids `factory.persistence` becoming a second factory, while keeping the session-pinning and preservation guarantees. Providers without support must reject persistence options before provisioning.

PR #572, opened 2026-09-13
agents: add sandbox environment configuration and thread bindings

quentinnippert:
## What changes

Move environment configuration into sandbox factories and bind conversations through typed parameters. Builds on #571; part of #373.

```ts
const host = await createSixb({
  // ...providers
  sandboxes: new VercelSandboxFactory({
    params: { clientId: param("string") },
    resolve: async ({ params, sixb }) => {
      const client = await sixb.objects(Client).get(params.clientId)
      const repositoryUrl = client?.properties.repositoryUrl
      if (typeof repositoryUrl !== "string" || !repositoryUrl.trim()) {
        throw new Error("Client repository is unavailable.")
      }
      return {
        source: { type: "git", url: repositoryUrl },
        setup: ["bun install"],
      }
    },
  }),
})

// On the application's authorized SDK instance:
await sixb.agent.threads.create({ sandbox: { clientId: "acme" } })
```

`Client` is an application ontology type with a `repositoryUrl` property.

## Design and boundaries

- **Factory:** common `source`, `setup`, `env`, and `network` contract; static configuration or typed `params` + `resolve`. No separate `agentWorkspace` configuration.
- **Thread:** optional, immutable parameters, validated with the existing Action parameter system. Stored as `sandbox_params`, not as a serialized sandbox or resolved environment.
- **Authority:** resolvers receive the execution-scoped SDK; saved parameters do not grant repository access.
- **Storage/API:** consistent bindings across memory, PostgreSQL, SQLite, HTTP and the regenerated client. Existing pending migrations are updated in place.
- **Execution:** thread creation neither resolves nor provisions compute. Persistent bindings require `resume`; bound runs/retries remain blocked until #577. Omitting `sandbox` preserves ephemeral behavior.
- **Documentation:** app-author examples in public docs; primitive internals in the sandbox README.

## Validation

- 205 targeted tests passed, including worker guards, API validation and SQLite reopen persistence.
- Source/test typechecks, lint and diff checks passed.
- Validation regression tests fail when the field validators are temporarily bypassed.
- Earlier full-suite run: 5,459 passed, 7 skipped; one unrelated CLI scaffold failure (`cp EEXIST`).
- No live cloud or PostgreSQL tests run for this revision.


demattosanthony:
Is there a way to avoid introducing `agentWorkspace` config and have this be part of the sandbox instead?

I like the Anthropic/OpenAI core concepts:

- Agent — model, prompts, tools, skills
- Environment — configuration for where the session runs
- Session — an agent working inside that environment
- Events

https://platform.claude.com/docs/en/managed-agents/overview#core-concepts

I feel like for us that maps to **Agent, Sandbox, Threads/Runs, and Events**.

In my head, the sandbox config describes how to set up the environment, and the thread gives it continuity across runs. 

Something like this for the dynamic case:

```ts
export const host = await createSixb({
  // ...

  sandboxes: new VercelSandboxFactory({
    params: {
      clientId: param("string"),
      branch: optional(param("string")),
    },

    // Check access on every run, including when restoring.
    authorize: async ({ params, sixb }) => {
      await requireRepositoryAccess(sixb, params.clientId)
    },

    // Set up a fresh sandbox once.
    setup: async ({ sandbox, params, sixb }) => {
      const client = await requireRepositoryAccess(sixb, params.clientId)

      await cloneRepository(sandbox, {
        url: client.repositoryUrl,
        branch: params.branch ?? client.defaultBranch,
      })

      await installDependencies(sandbox)
    },
  }),
})

const thread = await sixb.agent.threads.create({
  title: "Acme settings page",
  sandbox: {
    clientId: "acme",
    branch: "main",
  },
})
```

And the simple case stays simple, with no params or callbacks:

```ts
export const host = await createSixb({
  // ...

  sandboxes: new VercelSandboxFactory({
    source: {
      type: "git",
      url: "https://github.com/acme/app",
      revision: "main",
    },
    setup: ["bun install"],
  }),
})
```

Not tied to these exact API names.


But if you think `agentWorkspace` is best then lets do it

PR #577, opened 2026-09-14
agents: persist conversation workspaces across runs

quentinnippert:
## Summary

Slice 3 of #373, stacked on #572. Enable opt-in conversation workspaces that retain files and uncommitted work across runs.

```text
Thread workspace binding
  → Resolve recipe under current execution authority
  → First run: create({ persistence: { name } }) + clone + setup
    Later runs: resume(name, options)
  → Refresh environment, runtime files and network access
  → Agent works in the checkout
  → Clean up + confirm snapshot preservation
  → Finalize the run
```

## What's added

- Shared lifecycle guards and durable workspace state for memory, SQLite and PostgreSQL.
- Generation-checked recreation through the SDK, HTTP API and chat confirmation; generated client and concise usage docs.
- Unified sandbox creation and optional `resume()` capability; retention alone never enables a workspace.

## Safety choices

- Pin repository identity and reject stale owners to prevent conflicting workspace use.
- Block uncertain state rather than silently replace files. Recreation starts fresh without recovering uncommitted work or deleting the old sandbox.
- Automatic crash recovery remains deferred: provider operations can outlive a worker. Private Git authentication belongs to the next slice.
- Threads without workspaces, workflows and subagents remain ephemeral.

## Validation

- Current revision: 5,321 tests passed, 7 opt-in skips; 205 targeted tests, source/test typechecks and lint passed.
- Original slice: 62 PostgreSQL checks, builds, client generation and 56-package publication checks passed; those contracts are unchanged by this revision.
- No live Vercel resources used. Recovery UI has server-rendered coverage, not browser interaction tests.


demattosanthony:
The `AgentThreadRecord` storage type currently has:

```ts
readonly sandbox?: AgentThreadSandbox
readonly workspaceState?: AgentWorkspaceState
```

`sandbox` holds the parameters used to resolve the sandbox recipe, while `workspaceState` tracks that sandbox’s identity, status, and ownership. Would naming both around the sandbox make their relationship clearer?

```ts
/** Immutable application inputs used to resolve the sandbox recipe. */
readonly sandboxParams?: AgentThreadSandboxParams

/** Worker-managed lifecycle state of the thread's persistent sandbox. */
readonly sandboxState?: AgentThreadSandboxState
```

We currently open a workspace separately and pass it into the agent environment, which delegates saving back to it:

```ts
const workspace = await openAgentWorkspace({ thread, run, /* ... */ })

const environment = await createConversationAgentEnvironment({
  workspace,
  /* ... */
})
```

Could createConversationAgentEnvironment also own acquiring and saving that sandbox, with the lifecycle helper kept internal? That would give the worker one environment lifecycle to coordinate.

```ts
const environment = await createConversationAgentEnvironment({
  thread,
  run,
  /* ... */
})
```

It already prepares the sandbox for a run. With persistence, that lifecycle would become create or resume → prepare → execute → clean up → stop. The existing provider supplies the persistence operations:

```ts
await factory.create({ persistence: { name }, /* ... */ })
// or:
await factory.resume(name, sessionOptions)

// Before finalizing the run:
await drainSandboxOperations()
await cleanRunFiles(sandbox)
await sandbox.stop() // Confirms preservation.
```

The ownership checks and durable state transitions would still surround these operations. We could also store the persistent sandbox name directly: `generation` currently identifies a sandbox incarnation, rather than a version of its files, so a reset could simply assign a fresh name.

PR #600, opened 2026-09-17
core: manage vector profiles with atomic invalidation

quentinnippert:
## Summary

The existing vector API stores embeddings as ordinary numeric properties, without tracking which source values or model produced them. Updating a description can therefore leave its old embedding searchable.

Named vector profiles make embeddings derived search state: one vector per object and profile, stored separately from business properties. Each profile declares its sources and references a model registered in `models.embedding`.

```ts
search: {
  vectors: {
    content: { source: ["title", "description"], model: productEmbedding },
    context: { source: ["context"], model: productEmbedding },
  },
}
```

```ts
await sixb.objects(Product).byId(id).vector("content").index()
await sixb.objects(Product).query().vector("content", queryVector, { k: 10 }).list()
```

- **Safe writes:** `index()` reads sources and generates outside the transaction, then atomically rejects stale object or vector revisions. Reindexing leaves the object's version and events unchanged.
- **Fresh results:** edits and projections invalidate affected profiles in the object transaction; unrelated changes preserve them. Configuration changes exclude incompatible vectors.
- **Predictable search:** validated float32 vectors, bounded cosine top-k, authorization and filters before ranking, deterministic ties, and scores preserved through HTTP and the client.
- **Shared model configuration:** an embedding contract and catalogue support projects without language models.

## Scope

Slice 1 establishes the complete lifecycle in memory. PostgreSQL persistence/search and production qualification follow separately; PostgreSQL and SQLite currently reject named profiles at startup. No built-in embedding provider adapter is included.

This replaces `search.vector`, property `query.vector` and vector-query `propertyId`; only named profiles remain. Numeric arrays stay ordinary business properties. Regeneration is explicit: invalidated objects stay absent from vector results until `index()` runs. Automatic scheduling and discovery of profiles needing reindexing are deferred.

## Validation

- Full unit suite: 6,240 passed, 7 skipped, 0 failed.
- Full typecheck, Biome and client generation passed.
- Regression checks reject removed metadata and wire selectors; removing the guards makes the tests fail.


packages/core/src/objects/query/ir.ts
```diff
@@ -78,7 +78,12 @@ export interface ObjectQueryVector {
   kind: "vector"
   input: ObjectQuery
   vector: readonly number[]
-  propertyId: string
+  /** @deprecated Legacy numeric property queries. Prefer a named profile. */
+  propertyId?: string
+  profile?: string
+  /** Populated by canonical validation, never trusted from the wire. */
+  configuration?: string
+  source?: readonly string[]
```
demattosanthony: instead of deprecating should we clean break?

we didn't really have a system for using it properly before

packages/core/src/objects/vectors/handle.ts
```diff
@@ -0,0 +1,90 @@
...
+        ref,
+        profile: profileName,
+        configuration: vectorConfiguration(profileDefinition),
+        ...vectorSources(profileDefinition.source, row.properties),
+        expectedObject: { ref, exists: true, version: row.version, lastCommitId: row.lastCommitId },
+        expectedVectorCommitId: current?.lastCommitId ?? null,
+      }
+
+      // Network work never holds a storage transaction or retries an obsolete computation.
+      const result = await model.embed({ texts: [input.text] })
```
demattosanthony, line 68: should embedding calls follow the same operational pattern as language models?

- usage + costs/accounting
- limits

model.embed() just goes straight to the provider 

maybe fine to start with but could be good to have

quentinnippert: made some updates on the stack

PR #611, opened 2026-09-19
pg: prevent quadratic cardinality validation

quentinnippert:
## Summary

Large projections can trigger quadratic final cardinality validation: PostgreSQL underestimates relevant records in the mixed work table and repeatedly scans links for each scope.

- Copy effective cardinality records into a typed temporary table and analyze that focused population.
- Preserve duplicate detection and both expected/actual link-set checks.
- Skip validation when no cardinality work exists.
- Clean up temporary tables on success or transaction termination; avoid accumulating prepared statements per session.

## Performance

Same-fixture regression checks, counting link rows visited:

| Scenario | Before | After: verified upper bound |
|---|---:|---:|
| 5,000 scopes / 5,000 links | 25,000,000 | ≤ 40,000 |
| 100 scopes / 20,000 links | 2,000,000 | ≤ 80,400 |

With the fix, the real materializer committed **197,682 subscriptions and links in 189.57 s**; the final validation SELECT took **703 ms**. Synthetic data, PostgreSQL 17.10, Bun 1.4.2; SELECT timing excludes temporary-table preparation. This larger run is a scale check, not a matched before/after timing comparison.

## Design choices

- Analyzing the mixed table alone still fails sparse workloads; the dedicated table removes that sampling dependency.
- No persistent schema migration or global planner settings.
- Tradeoff: three additional SQL statements and a copy of relevant records per cardinality finalization. Remote latency for frequent small writes needs deployment-specific measurement.

## Validation

- **485 tests passed:** 353 PostgreSQL E2E, 27 PostgreSQL unit, 105 projection-worker unit.
- 13 new regressions cover dense/sparse work, low statistics precision, invalid links, rollback and cleanup. Restoring the old query fails the performance guards.
- Source/test typechecks, PostgreSQL build, Biome and diff checks passed.
- Separate worker experiment verified timeout rollback and explicit retry without duplicate publication.

## Scope

PostgreSQL final validation only. No SQLite changes; the full repository test suite was not run. Progress reporting and retry limits remain follow-up work.


quentinnippert:
@demattosanthony can you prioritize this one during your next review session pls ?

PR #612, opened 2026-09-19
storage: durable, swept file upload sessions for pg and sqlite

### LINKED ISSUE #279: Task: durable, swept file upload sessions for @sixb/pg and @sixb/sqlite
A staged upload spans four requests, and the `uploadId` that ties them together lives in the
memory of one process. `FileUploadSessionStore` is optional on `Storage` and only
`InMemoryStorage` implements it, so `@sixb/pg` and `@sixb/sqlite` both fall through:

```ts
// packages/server/src/routes/files.ts:46
const uploadSessions = sixb.storage.fileUploadSessions ?? new InMemoryFileUploadSessions()
```

Every production deployment takes that fallback. Which means the multi-request upload path —
the one built for large files — is the only part of the API that cannot run behind a load
balancer, while `docs/deployment/overview.md` says `sixb api` scales to many replicas.

## What breaks

```
POST /api/files/uploads          -> replica A   creates session upload_9f3c…
POST /uploads/upload_9f3c/parts/1 -> replica A   200, signed URL
PUT  <signed url>                 -> S3          part uploaded
POST /uploads/upload_9f3c/complete -> replica B   404 — B never saw this session
```

The user has uploaded 800 MB to S3 and cannot finish. The parts stay in the bucket, billed,
with no session left to abort them. A rolling restart does the same thing with one replica.

## What is missing

| | Today | Needed |
| --- | --- | --- |
| `FileUploadSessionStore` in `@sixb/pg` | — | table + migration |
| `FileUploadSessionStore` in `@sixb/sqlite` | — | table + migration |
| `cleanupExpired()` caller | only `InMemoryFileUploadSessions.create()`, lazily | a scheduled sweep |

The sweep is not optional extra credit. With a durable store and nothing reaping it, expired
sessions accumulate forever and their provider-side multipart uploads are never aborted —
the failure above, minus the 404, repeated for every abandoned upload.

## Scope

1. Implement the seven-method contract in both SQL providers:

   ```ts
   // packages/core/src/storage/file-upload-sessions/types.ts:53
   create · getForPrincipal · markUploaded · addSignedPart · complete · abort · cleanupExpired
   ```

   `InMemoryFileUploadSessions` is the reference for the state machine: `pending` →
   `markUploaded` stays non-terminal, `complete`/`abort` are terminal, terminal rows are kept
   for `DEFAULT_FILE_UPLOAD_TERMINAL_SESSION_TTL_MS` so a retried `complete` stays idempotent.

2. Call `cleanupExpired()` from `OntologyMaintenance`, alongside the outbox and source-row
   purges it already runs every 60s in the API role. A session whose provider upload is still
   open should be aborted through `blobStorage.abortUpload` before the row goes.

3. Drop the fallback in `files.ts` once every published provider implements the store, or keep
   it and warn — a silent fallback is what let this ship.

4. Document the `AbortIncompleteMultipartUpload` lifecycle rule as the second line of defence
   in `docs/infrastructure/overview.md`: a sweep that never runs because the API is down still
   leaves parts behind, and only the bucket can clean those.

## Notes

- `@sixb/core/testing` already ships shared contract suites (`ontology-storage-contract`,
  `auth-storage-contract`, `agent-storage-contract`, …). Add a `file-upload-session-storage`
  one and run it from both providers, rather than writing the tests twice.
- Single-request `POST /api/files` is unaffected — it opens no session.
- 0.1.0 ships this as a documented limit in `docs/server/overview.md`.

yashksaini-coder:
Closes #279.

Staged uploads (`POST /api/files/uploads` → parts → complete) kept their session in the serving
process: `@sixb/pg` and `@sixb/sqlite` had no `fileUploadSessions`, so `files.ts` fell back to
`InMemoryFileUploadSessions`. Behind a load balancer, or across a restart, `complete` hit a replica
that had never seen the session, and the provider-side multipart parts stayed in the bucket with
nothing left to abort them.

## What changes

- **One state machine, three stores.** `DurableFileUploadSessions`
  (`@sixb/core/internal/file-upload-session-storage-provider`) owns every transition and retention
  rule over a narrow persistence backend, the same shape as `DurableConnectorConnectionStorage`.
  `InMemoryFileUploadSessions` is now that class over a `Map`; `SqliteFileUploadSessionStorage` and
  `PgFileUploadSessionStorage` implement row access only. Migration `041-file-upload-sessions` on both.
- **Abandoned uploads are never silently deleted** (option (b) from the thread). An expired
  `pending` session that still holds a `providerUpload` is only listed by the new
  `listAbandoned({ projectId, now, limit })`; `cleanupExpired` and `getForPrincipal` never delete it.
  Retention is one core-derived value, `fileUploadSessionReapAt`, persisted as `reap_at`, so SQL
  deletes `reap_at <= now` and never restates the rule.
- **Sweep.** `OntologyMaintenance` gains a third cleanup domain: list abandoned → `abortUpload`
  → `abort` (terminal) → reaped after the terminal TTL. It leaves a session alone until 5 minutes
  past expiry (a request that passed the expiry gate may still be completing provider I/O, and the
  API clock can lead the database's). A failed provider abort is retried for 15 minutes past expiry,
  then the row is retired anyway and the failure is still reported, so an un-abortable row cannot
  hold a slot in every pass's `cleanupLimit`. A row another replica retired first counts as done.
- **Fallback kept, no longer silent.** A custom `Storage` without the namespace still gets the
  in-memory store, with a startup `console.warn`; that instance drops abandoned sessions on
  `create` since no maintenance pass sees it.
- Docs: the pre-0.1 limit note in `docs/server/overview.md` is replaced; `docs/infrastructure/overview.md`
  documents `AbortIncompleteMultipartUpload` as the second line of defence.

## Decisions worth a look

1. **Expiry is enforced at the gate, not in transitions.** Every route calls `getForPrincipal` first,
   then does blob I/O, then transitions. Re-checking expiry in `complete` would 410 a request whose
   provider completion finished a second after `expiresAt`, leaving a completed blob unreferenced and
   the row looking abandoned. This matches the previous in-memory behaviour; it is now written down in
   the interface doc and pinned by the contract suite.
2. **`getForPrincipal` no longer deletes on read.** It still reports `expired`/`not_found`; deletion
   belongs to the sweep. Otherwise a client retrying after expiry would delete an abandoned row
   before anything aborted its provider upload.
3. **Lookups are by `id`, not `(project_id, id)`.** The store's methods take only `uploadId`, and ids
   are server-minted 128-bit (`upload_<uuid>`), so `id` is the primary key; `project_id` is stored and
   `listAbandoned` is project-scoped. Widening every signature to carry `projectId` felt out of scope —
   happy to do it if you'd rather.
4. **Concurrency.** Pg transitions `SELECT … FOR UPDATE`. SQLite needed a per-store promise chain:
   `runImmediateTransactionAsync` *joins* an already-open transaction instead of waiting, so two
   concurrent `addSignedPart` calls interleaved their read-modify-write and dropped a part. The
   contract suite's concurrency case fails without the chain.
5. **#597 compatibility.** `providerUpload` and `signedParts` are stored as JSON and revived with only
   `expiresAt` parsed, so #597's `partReceipt` / optional `etag` round-trip without a migration (the
   contract suite round-trips an unknown provider field).

**Migration number:** rebased onto `main` at `f41628c6`; now `041` on both providers (038–040 were taken upstream).

Not included: a `fileUploadSessionsDeleted` counter in the maintenance snapshot — it would touch the
status schema, OpenAPI and the generated client. Easy follow-up if wanted.

## Verification

- `file-upload-session-storage-contract` (12 cases) runs against in-memory, SQLite, and Postgres.
- Sweep: 9 cases in `file-upload-session-sweep.test.ts` with a fake direct-upload blob storage.
- Every guard was removed once to watch its test fail: signed-part sort, transaction serialization
  (in-memory and SQLite), gate-only expiry, abandoned rows unreapable, abandoned ordering, project
  filter (in-memory and SQLite), sweep delay, grace retry, give-up, race tolerance, cleanup limit,
  unswept drop.
- `bun run check`, `bun run typecheck`, `bun run build`, `bun run test:publish` (57 packages),
  `bun run generate:client` (no drift) — all clean.
- `bun test`: the only failures are the 55 `packages/cli` tests that fail identically on `main`
  on a Bun 1.4.0 machine (`[Sixb] Bun 1.4.2 or newer is required`); same set by name before and after.
- `@sixb/pg` e2e (real Postgres 17 in Docker), after rebasing onto `f41628c6`: **376 pass / 0 fail** across 32 files, including the upload-session contract suite and all three migration-history fixtures with `041`.


pm25coder:
Came here from the write-up on the Floodsub deletion, which named this PR as the heavy lift. The part I wanted to check in the code before believing "durable … spans API replicas" is the sweep's race, so I read the four pieces that decide it and measured the SQLite side locally.

**Does the delete need to be conditional?**

No — and the reason is that the predicate is re-evaluated once the lock is granted, so it cannot delete a row a transition has just moved. I drove exactly your statement (`storage/sqlite/src/file-upload-session-storage.ts:115`) from a second connection while a writer held `BEGIN IMMEDIATE` and pushed `reap_at` to now+900s: the sweeper blocked 1151 ms, then deleted **0** rows, and the row survived. On SQLite 3.47.1 / CPython 3.13 — that is the engine's semantics for the same SQL shape, not your code (no bun on this box, so nothing here is your suite running).

**Where the guarantee actually lives is four brackets**, and that is what I would call out in a write-up rather than the SQL:

1. `ABANDONED_UPLOAD_SWEEP_DELAY_MS = 5 * 60_000` (`packages/core/src/maintenance/ontology-maintenance.ts:25`) — the sweep only looks at sessions that expired five minutes ago (`:209-211`), so a request that passed the expiry gate and is still finishing blob I/O was never a candidate.
2. The decision itself is read unlocked — `listAbandoned` goes through `backend.read` (`packages/core/src/storage/file-upload-sessions/provider.ts:135`), predicate at `storage/sqlite/src/file-upload-session-storage.ts:104` and `storage/pg/src/pg-file-upload-session-storage.ts:95`. The decision can be stale by construction; that bracket is the one doing the work.
3. The transition re-reads under a lock and refuses non-pending rows (`provider.ts:148-155`, `FOR UPDATE` at `pg-file-upload-session-storage.ts:66`, `BEGIN IMMEDIATE` at `storage/sqlite/src/transactions.ts:83`).
4. The delete re-evaluates under the same write lock (the measurement above).

**SQLite vs Postgres is not "same query, different behavior" on the delete** — the predicate is identical — it is different in who waits: `BEGIN IMMEDIATE` holds the whole write lock for the transaction (plus the in-process `tail` chain at `file-upload-session-storage.ts:23-37`, and `busy_timeout = 5000` at `transactions.ts:33`), while PG locks a single row from `FOR UPDATE` to COMMIT and leaves the rest of the table writable. Two sweepers racing: the loser blocked 539 ms and then deleted 0 — a lost race is a quiet no-op, not an error. One consequence worth naming: `deleteReapable` returns a count (`provider.ts:138-141`), so a caller cannot tell "0 because another instance swept it" from "0 because nothing was ready".

**The one thing I would change is in the accounting, not the concurrency.** `retireAbandonedUpload` treats `already_aborted` as done (`ontology-maintenance.ts:250-253`) but re-throws `already_completed` (`provider.ts:153-155`). The outcome is right — refusing to abort a completed provider upload is correct, its parts are already assembled — but the pass records a failure, `consecutiveFailures` increases, and two in a row flip `getOperationalStatus()` to `degraded` (`DEGRADED_FAILURE_THRESHOLD = 2`). Reachability is narrow: the route's `getForPrincipal` throws `expired` on any row the sweep could list, so it takes a provider-side complete that runs more than five minutes past expiry. Is landing in `degraded` intended there, or should the completed case mean "nothing to do" the way the aborted one does?

The test that would settle it: `packages/core/tests/file-upload-session-sweep.test.ts:97` pins the aborted race ("another instance retired first"); the completed outcome — the only branch that throws — has no case.

quentinnippert:
@yashksaini-coder @pm25coder Thanks for your work on this. For future PRs and comments, could we keep things a bit more concise and take the time to fully understand the changes before submitting or reviewing them? It’ll make collaboration easier for everyone.

